PR #1, opened 2017-09-01
Fix userfunc issue

boldowa:
#  bugfix
- [SMWC](https://smwc.me/1404712)
- [My repository](https://github.com/boldowa/Asar/issues/2)

I Added function initialization at reset.


RPGHacker:
Hey! Thanks for the fix! 😉 
I currently still have a bunch of local changes to Asar on my system that I haven't commited yet (in fact, this repository currently is stil very incomplete, the CMake files aren't even fully usable yet). Once I'm done with my local changes and have commited everything, I will verify this fix and then merge it into my branch. Appreciate the help a lot! 👍

boldowa:
😉

I can provide the environment I used to check bug fixes. Please contact me if you need.

RPGHacker:
I'll take a look at this one now. Can you provide the test environment you've mentioend?

boldowa:
ok

[velify_asar](https://github.com/boldowa/velify_asar)

I uploaded my test environment.
Please check it.

## by the way...
dll (or so) seems to be incompatible between versions 1.40 and 1.50, is this a specification?

This can be confirmed by rewriting the version information of line 6 of CMakeList.txt to 1.40.

RPGHacker:
Thanks for the test cases, I've verified the bug and merged your fix (and then moved it to a slightly more convenient place). You can verify the fix for yourself if you want.

As for Asar DLL version incompatibility, I think this is by design, though I haven't tested enough with it to know how well everything works. Logically speaking, an older app should be compatible with any newer DLL and not require a rebuild, whereas a new app doesn't necessarily have to be compatible with older DLLs. Don't know if it actually works like this, though, as I haven't tested this too much yet.

PR #3, opened 2017-09-25
Add features from asar-tc ("import" and "undef" commands)

zetaPRIME:
`import`: acts as an incsrc from a target-relative search directory (defaults to `./lib/`)
`undef`: explicitly deletes a define

### DIFF
--- a/src/asar/asar.h
+++ b/src/asar/asar.h
@@ -9,7 +9,6 @@
 #pragma once
 #define Asar
 
-#include <libgen.h>
 #include "autoarray.h"
 #include "scapegoat.hpp"
 #include "libstr.h"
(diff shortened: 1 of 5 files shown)


zetaPRIME:
...oops, seems the libgen dependency breaks compilation on windows outside of a MinGW environment... will fix as soon as I figure out the best way to do so

RPGHacker:
Hey there! Thanks for the contribution! :)

I had actually planned to work on include directories as one of my next features, but didn't get to it yet, so I guess I can make good use of this pull request, haha. Will probably change one or two things here and there, but should give me a good head-start.

Didn't even think about undef, but that seems like quite a useful feature, too.

It will take a while before I can take a closer look at the pull request and merge it, because I can barely find time to actively work on Asar right now, but the help is certainly appreciated. 👍

PR #5, opened 2018-02-09
Make test-suite compile under linux. 

Horrowind:
They still fail, but this will be in a different commit.


RPGHacker:
Thanks, definitely planned to do that, never got around to it yet. Will definitely merge it once I get around to it (will probably take a while, though).

RPGHacker:
Help! How the frick do I do things on Linux?!?

I've set up an Ubuntu VM a few days ago (mainly because I've had at least some experience with Ubuntu already and kinda liked it), cloned the repository via git, locally merged the changes from this pull request, installed cmake to configure the project and have since tried three different IDEs to try and get poop done, but have horribly failed. So far I've tried Code::Blocks, Eclipse CDT and Codelite. Eclipse CDT was just... what... I was completely lost and didn't get anything. It took me forever to even get anything to compile and launch, only to find out that it wouldn't find any symbols and thus I could only debug on an assembly level. Code::Blocks was slightly more intuitive and I've worked with it in the past, but have forgotten almost everything (I remember disliking aspects of it in the past already). At least building everything was easy in it, but I can't debug in it, the debugger just always ignores my breakpoints and stuff. I've already googled for a solution for hours and have tried the most common solutions (make sure -g is enabled, make sure "strip symbols" is disabled etc.). Finally I've tried Codelite, which was my favorite so far (because it is very similar to Visual Studio, which I'm used to) and again could build and launch the application without problems, but again couldn't debug anything because the debugger would ignore my breakpoints (I've looked at the debugger log and there was nothing suspicious there - it correctly set all of my breakpoints, but still ignored them). Again I've made sure -g is set, and again I've googled for hours and didn't find a single working solution so far.

Just... how do you do stuff on Linux? Getting something as essential as debugging your application to work should not be this difficult. Do people not debug their applications at all, or am I just a complete idiot, failing at the most basic tasks on Linux? 😞

RPGHacker:
Alright, problem solved.

`cmake -G"CodeLite - Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ../src`

Need to set the CMAKE_BUILD_TYPE to Debug for debugging to work. Don't quite understand why, since the -g option was specified even without doing this, but oh well. At least I can finally get something done now.

PR #27, opened 2018-03-13
Bugfixes and some new features and stuff

### LINKED ISSUE #22: Add a command to verify the title of the output ROM
Add a command to verify the title of the output ROM and throw an error if it doesn't match. for example:
`expect "SUPER MARIOWORLD     "`
This allows us to make sure that a certain patch is only used with a certain ROM.

Apparently a function like this has existed in Asar at some point, but was removed for an unknown reason.

### LINKED ISSUE #25: opencachedfile() might fail when opening > 16 files in a single patch
Don't have the time to actually test it right now, but adding the issue so that I don't forget to check later.

Looking at the code of opencachedfile(), it seems as though the function isn't very smart and will most likely fail (or even crash) if we attempt to open a total of more than 16 different files in a single patch. While not likely to be done, it can technically happen and should be fixed if that's the case.

### LINKED ISSUE #29: Add pushbase/pullbase commands
Requested by Ladida. Presumeably they should work just like pushpc/pullpc, except on the base instead of the pc.

randomdude999:
* Add a .rc file for the DLL (tested)
* Add version info to the main exe's .rc file (tested)
* Fix a bug with ROM title reporting
* Add `expecttitle` (it is case-sensitive, but that may be a good thing) (closes #22) (tested, but not very thoroughly)
* Add some kind of pushbase/pullbase prototypes (untested, but if it works it'll close #29)
* Allow escaping ! using \\! (tested)
* Fix the file-reading functions crashing after reading 16 different files. (closes #25) (tested)

Possibly todo: add a way to set the `ignoretitleerrors` flag from the DLL interface


RPGHacker:
Small suggestion: Maybe rewrite the expecttitle error and warning messages a little so that instead of
"ROM title is incorrect"
they say something like
"ROM title is incorrect. Expected '%s', got '%s'"
I'm not sure if error() and warn() already support printf formatting, but you could just wrap everything in Asar's string class and then use + to concat everything.
```
warn(S "ROM title is incorrect. Expected '" + expected_title + "', got '" + actual_title + "'");
```

randomdude999:
They don't support formatting, but yeah, I could just concatenate them.

randomdude999:
The last commit should close #25.

PR #28, opened 2018-03-13
Absolute path include / CMakeFile

boldowa:
This pull request includes the following changes.

# Absolute path include feature

I noticed that ASAR can't include with an absolute path.

Since there was a problem at the time of tool creation, I changed it so that it can be included in absolute path.

# CMakeFile options

Since it is convenient to be able to limit the target of compilation, we added the following three options.

- ASAR_GEN_TEST
- ASAR_GEN_EXE
- ASAR_GEN_DLL

In particular, I think that the test code should be excluded from compilation because it can't be built in the MinGW environment.

For every option, the default value is ON (generate code).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,10 @@
 cmake_minimum_required(VERSION 3.9.0)
 
+option(ASAR_GEN_TEST "Generate tests" ON)
+
 add_subdirectory(asar)
 
-add_subdirectory(asar-app-test)
-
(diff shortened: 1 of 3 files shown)


RPGHacker:
Thanks!
I'm not currently at home, so I will merge later.

For the absolute paths, it might be a good idea to move the check into a function to reduce copy-paste code.

```
bool is_absolute_path(const char* path)
{
        return (...) ;
}
```

PR #35, opened 2018-03-19
I made another pull request

### LINKED ISSUE #6: Proposal: Remove xkas support.
From the manual:
> New features introduced into Asar since then might not throw warnings when attempted to use in xkas compatibility mode and old xkas patches might not assemble correctly with Asar anymore, even when xkas compatibility mode is used.

Maybe we should remove xkas support altogether? (Probably after a community discussion?)

### LINKED ISSUE #15: Make namespaces stackable
Namesspaces should be stackable, like in C++. Maybe add a "namespace exit" command which only leaves the current namespace rather than immediately returning to the global namespace.

Code example
```
namespace first

    namespace second
        Table:
            db Table ; first_second_table
    namespace exit

    Table:
        db second_Table ; first_second_Table
        db Table ; first_Table

namespace exit

Table:
    db first_second_Table ; first_second_Table
    db first_Table ; first_Table
    db Table ; Table
```

### LINKED ISSUE #32: Add an option to ignore bank crossings everywhere
Requested by Ladida: add a feature to make Asar ignore bank crossings everywhere (instead of just certain incbin commands). Either via a command line argument or via a new command.  Not sure which one would be better, but I think a new command might be, because you could disable the check only for a certain section of code or something like that.

### LINKED ISSUE #36: Continous Integration
We could use it to automatically build release files (using Appveyor) and also to verify that pull requests don't break anything (probably using Travis, but i think Appveyor might work too and IMO it would be better to use only one).

### LINKED ISSUE #37: Add a way to patch a file without calling asar_reset
This could be done by adding another flag to asar_patch_ex, maybe called "should_reset" or something.
(requested by TheBiob on discord)

randomdude999:
(this is my third one already)
* Add a warning for using ;@xkas or warn xkas (should close #6 for now)
* Add nested namespaces (closes #15)
* Add continous integration in the form of Appveyor and Travis CI. (closes #36) RPG Hacker, what you have to do when merging this is:
  * Go to https://travis-ci.org, log in with your github account, click the checkbox thingy next to RPGHacker/asar
  * Go to https://ci.appveyor.com, click on login with github, click new project, select github, select asar
* Allow disabling bank cross errors with "warn bankcross off" (it's not technically a warning but whatever) (closes #32)
* Update the python wrapper with asar_patch_ex
* Add a way to not reset stuff with asar_patch_ex (closes #37)
* Okay i'm not even trying with the include guards anymore

Right now everything is untested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,7 @@ install:
   # Download a recent version of CMake
   - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
   - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
-script: cmake && make
+script:
+  - cd src && mkdir build && cd build
+  - cmake .. && make
 # TODO: actually test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,19 @@ install:
   # Download a recent version of CMake
   - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
   - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
+  # make links for the newer gcc
+  - ln -s /usr/bin/gcc /usr/local/bin/gcc
+  - ln -s /usr/bin/g++ /usr/local/bin/g++
 script:
   - cd src && mkdir build && cd build
   - cmake -DCMAKE_BUILD_TYPE=Debug ..
   - make
+# I have no idea what i'm doing but w/e
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-6
+      - g++-6
 # TODO: actually test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ install:
   - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
   - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
   # make links for the newer gcc
-  - ln -s /usr/bin/gcc-6 /usr/local/bin/gcc
-  - ln -s /usr/bin/g++-6 /usr/local/bin/g++
+  # - ln -s /usr/bin/gcc-6 /usr/local/bin/gcc
+  # - ln -s /usr/bin/g++-6 /usr/local/bin/g++
   - export CC=/usr/bin/gcc-6
   - export CXX=/usr/bin/g++-6
   - gcc -v && g++ -v && cmake --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: cpp
+install:
+  # Download a recent version of CMake
+  - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
+  - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
+script: cmake && make
+# TODO: actually test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,8 @@ install:
   - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
   - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
   # make links for the newer gcc
-  # - ln -s /usr/bin/gcc-6 /usr/local/bin/gcc
-  # - ln -s /usr/bin/g++-6 /usr/local/bin/g++
   - export CC=/usr/bin/gcc-6
   - export CXX=/usr/bin/g++-6
-  - gcc -v && g++ -v && cmake --version
 script:
   - cd src && mkdir build && cd build
   - cmake -DCMAKE_BUILD_TYPE=Debug ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 language: cpp
+sudo: required
 install:
   # Download a recent version of CMake
   - wget -O cmake.sh https://cmake.org/files/v3.10/cmake-3.10.0-rc1-Linux-x86_64.sh 
   - sudo sh cmake.sh --skip-license --exclude-subdir --prefix=/usr/local
   # make links for the newer gcc
-  - ln -s /usr/bin/gcc /usr/local/bin/gcc
-  - ln -s /usr/bin/g++ /usr/local/bin/g++
+  - ln -s /usr/bin/gcc-6 /usr/local/bin/gcc
+  - ln -s /usr/bin/g++-6 /usr/local/bin/g++
+  - export CC=/usr/bin/gcc-6
+  - export CXX=/usr/bin/g++-6
+  - gcc -v && g++ -v && cmake --version
 script:
   - cd src && mkdir build && cd build
   - cmake -DCMAKE_BUILD_TYPE=Debug ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,9 @@ install:
   - export CC=/usr/bin/gcc-6
   - export CXX=/usr/bin/g++-6
 script:
-  - cd src && mkdir build && cd build
-  - cmake -DCMAKE_BUILD_TYPE=Debug ..
+  - mkdir src/build && cd src/build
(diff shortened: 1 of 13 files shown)


RPGHacker:
A few suggestions:
* interface-cli.cpp in main() also contains the following xkas compatibility code:
`if (!strncasecmp(myname, "xkas", strlen("xkas"))) errloc=stdout;`
That should also print the xkas deprecation warning.
* I would also change the warning slightly to mention "xkas support is being deprecated and will be removed in a future version of Asar".

I think include guards are indeed somewhat tough to implement since you technically need to store the full paths of files being assembled somewhere, and it's possible that only the filesystem abstraction knows about full paths. It **might** be possible that `thisfilename` contains the full path to the current file, but I'm not sure about it. Probably nees some testing.

RPGHacker:
Might make sense to change the "Send this patch to the maintainer so he can debug it" entirely. Thinking of something like
`Please create an issue on the official GitHub repository and attach a patch which reproduces the error.`
Also I wouldn't remove the Alcaro credit entirely, firstly because the license mentions having to credit the author (and having his name hardcoded into the application is a convenient way of assuring this), and secondly it feels kinda "mean" to remove his name. Instead, I would just slightly change the message to read something like
`, developed by Alcaro, maintained by RPG Hacker` or `, developed by Alcaro, maintained by Asar devs`. Probably the latter, since it leaves the possibility open of someone else maintaining Asar in the future without the output having to be adjusted. We could also add another print after that with the GitHub link (https://github.com/RPGHacker/asar). So the entire message would read something like:

```
Asar v1.60, developed by Alcaro, maintained by Asar devs.
Source code: https://github.com/RPGHacker/asar
```

randomdude999:
yeah, that's a better idea.

Alcaro:
Personally I'd go for some variant of 'originally by Alcaro', but that works too.

(full disclosure: I'm biased)

RPGHacker:
HOLY POOP! Are you THE Alcaro? The legendary creator of the Asar assembler himself? 😲
What an honor! I've never expected meeting you here!

<sup><sup>Yeah, why not, let's go with "originally", that's definitely very unmistakable.</sup></sup>

randomdude999:
About the bank crossing by alcaro on discord:
>I take no responsibility if that one breaks the label sizer or freespace finder, they kinda expect bank crossings to not be allowed.
>or it could end up working, who knows

Alcaro:
> HOLY POOP! Are you THE Alcaro? The legendary creator of the Asar assembler himself?

:poop:

randomdude999:
nvm include paths are still quite broken

RPGHacker:
Regarding the Appveyor/Travis stuff: do I need to do anything else after merging this pull request? (As in setting up an account or configuring stuff), or will it already work with whatever you've set up so far?
Also another question: is it possible to configure the build server so that it also builds Asar for other platforms (specifically Linux) all the time? Because that's something I'd consider genuinely useful. Being a Windows guy, I of course build the Linux version rarely, so it's easy for problems to slip through there.

randomdude999:
Travis already builds on linux, i just don't know a good way to export the artifacts. There's a whole list of the ways it can upload stuff [here](https://docs.travis-ci.com/user/deployment/), but i don't think any of them would be viable for asar (unless you happen to have amazon s3 or something).

RPGHacker:
I personally don't care too much about binaries (from what I understand of Linux, distributing binaries doesn't make too much sense there, anyways). As long as it performs the tests on multiple platforms, I already see that as a gain.

randomdude999:
In that case, just remember that Travis builds on linux and Appveyor builds on windows.

RPGHacker:
Gotcha. So I take it you can only ever have one of those at a time? (Or at least have GitHub only display the state of one of them?)

randomdude999:
No, it displays the state of both: check my fork's commits (https://github.com/randomdude999/asar/commits/master). If even one of them failed, it shows an X. Clicking on it shows more details about which one failed. For example, click on the second one marked X ("maybe i got the config right this time"). It worked fine in Travis but failed in Appveyor.

RPGHacker:
That looks very practical. Thanks for the explanation.

RPGHacker:
I think it might be a good idea to make separate branches for each new feature going forward. With how big this pull request has become already, merging it will become a bit of a challenge and quite time consuming. Multiple branches will also make it easier to decide on a per-feature basis on which pull requests to accept or not accept.

RPGHacker:
I've went with a slightly different approach to fix the include path problem. I don't really like the idea of making our include paths global and copy-pasting and/or moving filesystem functions out of it, so I've solved the problem directly in there. I believe this to be the cleaner, but also more future-proof solution, as it would work better with virtual files/in-memory files, should we ever implement them.

I suppose this means that some of your commits here need to be undone so that I can merge the pull request later on? (Specifically the stuff on the include path problem). Don't know if there is an easy way to do this in GitHub (something like a subtractive merge?).

randomdude999:
I can just `git revert` the fix commits.

RPGHacker:
Alright, that sounds good (hope there's nothing else useful in those commits).
Also I commited the test I used to test if everything works, in case you want to check for yourself. I've tested it in both the test suite, as well as via the standalone executable.

RPGHacker:
Small request for the future: before creating pull requests, please locally test your changes more thouroughly, preferably by using breakpoints in your code and by also including a new test to check in (when adding new features or changing existing features). When merging many of your pull requests so far, I came across a few oversights I had to fix first that would have likely been caught easily via a test, that's why we have the tests. In the case of this pull request, for example, the new namespace features don't work. See issue #15 for some sample code that can easily be used as a test and that throws an error. In this particular case, you forgot an else in assembleblock.cpp
`if (!stricmp(par, "off")) namespace_list.reset();`
(Might have other errors as well, but I just got started on the merge, so can't say for sure). I will fix this one myself locally, since I've already started on the merge, basically just mentioning it for future pull requests)

Also hope you don't take it personal or anything, because I'm really thankful for all the help. Had to fix stuff in pull requests by other people as well, but since you're contributing so much, I wanted to mention it because it's generally a good idea to work together with some more coordination. 😉

RPGHacker:
(Btw., I think I'm slightly changing the namespace implementation - playing around with it right now, I realize that the solution I proposed in the issue might break some backwards compatibility, so I think instead of doing it like this, I will add a compiler setting to either enable or disable nested namespaces)

randomdude999:
IMO a compiler flag for it is really not the best idea - then you'd have to have 2 versions of Asar downloaded, one with nested namespace support and one without, and it wouldn't be obvious which one you should use for which patch. If you are talking about `namespace exit` then yeah, IMO it and `namespace off` are kinda bad solutions. We could have `namespace_exit` and `namespace_off`, which would do the same thing except not prevent you from using them as namespace names. Then we could also deprecate `namespace off`, throwing a warning and saying `namespace_off` should be used instead.

RPGHacker:
I worded it badly. When I said "compiler setting", I really meant "compatibility setting" (that's what I called it in the manual). Wrote "compiler" because I still incorrectly refer to Asar as a compiler instead of an assembler in my head.

Basically, I already implemented the change locally and did it like this:

```
;@00 00
;@00 00
;@00 00
;@00 00
;@06 80 00
;@04 80 00
;@06 80 00
;@02 80 00
;@04 80 00
;@06 80 00
;@00 80 00
;@02 80 00
;@04 80 00
;@06 80 00
;@00 00
;@26 80 00



org $008000


namespace nested on


Main:
	db $00,$00

namespace First
{
	Main:
		db $00,$00
		
	namespace Second
	{
		Main:
			db $00,$00
			
			namespace Third
			{
				Main:
					db $00,$00
				
					dl Main
			
			}			
			namespace off
			
		dl Main
		dl Third_Main
	}	
	namespace off
		
	dl Main
	dl Second_Main
	dl Second_Third_Main
}
namespace off
		
dl Main
dl First_Main
dl First_Second_Main
dl First_Second_Third_Main


namespace nested off


namespace First
{
	namespace Second
	{
		namespace Third
		{		
			Main2:
				db $00,$00
	
		}
	}
}
namespace off


dl Third_Main2
```

This test succeeds with my local changes. All old patches using namespaces should still behave identically like this.

randomdude999:
That looks like a good solution.

RPGHacker:
("pragma" might have been a better word. Also note that the brackets in the example are optional and I put them there just for readability).

RPGHacker:
Note that after the merge, asar-app-test fails on one test, since xkas compatibility now throws warnings and the app test suite doesn't really differentiate between warnings and errors yet, so it treats the test as failing. Basically, we could just parse out warnings, but I don't really care enough right now to fix it.
(The DLL test suite still succeeds on all tests, since it treats warnings and errors separately)

PR #40, opened 2018-03-25
Update READMEs

### LINKED ISSUE #33: Have a proper README.md
Current file is just a leftover of old manual.txt. Have a proper Readme that doubles as a both a, proper overview for the GitHub page and a readme to bundle with future releases of Asar.

Things that should go into the Readme:
* Very basic, simple instructions for patch users how don't need to care about programming stuff ("double click Asar and it'll ask you for file names")
* Reference to the manual (ideally both the local and the online version) for patch programmers
* Shortly mention the Asar DLL and that its documentation can mostly be found in the headers of the bindings
* Reference to the changelog (ideally both the local and online version)

That's about it I think.

randomdude999:
I'm not very good at writing READMEs, but I think these will do for a start. Closes #33.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,47 @@
-Known bugs:
-- Due to how my define parser works, Label: !Define = Value will create error messages. Workaround:
-  Label: : !Define = Value. However, Label: Anotherlabel = Value will work.
-- else is treated as elseif 1 in all contexts, so attaching two else commmands to the same if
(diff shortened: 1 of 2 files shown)


RPGHacker:
It was actually a conscious decision to have to separate readme files, so that one can remain here on GitHub and contain information for tool programmers only, whereas the other one can be distributed with Asar and contain end user information. I believe it's actually better this way. See discussion in issue #33.

PR #43, opened 2018-03-29
Add commandline defines

### LINKED ISSUE #11: Allow to set defines via command line
Add support to set defines via command line, like in C compilers etc. Maybe also add support to set defines via a TXT file Asar automatically searches for and includes if it exists.

randomdude999:
Also make the python wrapper actually functional lol
Closes #11


RPGHacker:
`if (!defines.exists(key)) defines.create(key) = value;`
Maybe it would be wiser to have Asar throw an error here when a define already exists? Silently dropping those could lead to the rare (but very nasty) situation where someone accidentally passes a define to the application twice and then wonders why changing the second one doesn't seem to work.

randomdude999:
Yeah, that seems like a good idea.

randomdude999:
Actually, should it be necessary to add the error to the DLL interface too, or would just documenting that the last value is going to take effect be enough?

RPGHacker:
I've just commited some functionality to the master branch which adds functionality to Asar to read include search paths from a file called stdincludes.txt and defines from a file called stddefines.txt located next to the Asar executable (the test suite also generates files for that, it seemed like the most convenient way of getting tests for it working reliably). The stdincludes.txt part is fully implemented already, but the stddefines.txt part is just a stub right now since it heavily depends on this pull request. Do you want to take a stab at it? If so, you can try merging the master branch into this pull request's branch and then CTRL + F for a function called parse_std_defines() in main.cpp.

If you want to attempt it, make sure to also add the details to the manual (I've already added a stub for it, just CTRL + F for #TODO). While at it, you could also add the new parameters I just added on main to the Python wrapper.

The format of stddefines.txt isn't fully set in stone yet, which means the details would be up to you. Currently, the test suite generates a file like this:

```
!stddefined=1
 stddefined2=1
stddefined3
stddefined4 = 1 
```

On the left side (as in, before the = ), I think it only makes sense to ignore all whitespace and ignore a leading ! if present. Not sure how to handle whitespace on the right side, though, since the values of defines are allowed to contain whitespace. I have a few possible ideas for that:
* Just take everything right of the = verbatim, including all whitespaces
* Ignore leading and trailing whitespaces, but keep all other whitespaces
* Ignore leading and trailing whitespaces, unless the value is wrapped in double quotes, just like how Asar does it

The third solution might be the "best" soluiton here because the define format would be almost identical to defines in Asar, but it would require some extra effort (and complicate the code). The first solution would probably be the second-best one (or best one in terms of simplicity).

Anyways, let me know if you want to give it a try. If not, that's of course no problem and I'll attempt it myself after merging this pull request.

randomdude999:
Third solution shouldn't actually be too difficult, given that the current define parsing code already does that. I think it would be easier if you just merged this PR, i'll just make a new one.

RPGHacker:
Alright, I can do that. Need some time, though. I'll see if I can get to it later today.

randomdude999:
Actually nvm I got it merged myself. I'll see if i can get this working tonight

randomdude999:
Also why on earth are file-reading routines in libstr.cpp

RPGHacker:
Alright, thanks! 👍

Also no idea, I've stopped wondering about stuff in Asar long ago. 😂
Probably because it treats input files as text files and automatically skips \r for cross-platform compatibility.

Thinking about it... I've just realised that one version of readfile() skips \r for cross-platform compatibility, whereas the other version doesn't and is used by incbin. I was almost about to panic, because I thought incbin could actually be using binary data with all \r skipped, but luckily that isn't the case.

randomdude999:
okay i'm actually testing this right now

RPGHacker:
Should I already merge this one? Or do you want to fix the segfault (the one which causes Travis to fail) and add the new parameters to the python binding first?

randomdude999:
I'll work on it a bit more. I don't have a proper linux debugging environment set up right now so it'll be a bit difficult to debug the segfault, but I'll see what I can do. Also you should test whether the quotes work properly (!define = "hello").

RPGHacker:
I can take a look at the Linux part then (or, if you prefer, you can setup a VM with a Linux distribution, which should make debugging easier).

Alright, I'll see if can somehow add something to the test to verify whether quoted defines are working correctly.

randomdude999:
No, I mean that I have a linux machine available, it just doesn't have any IDE or stuff. Eh, gdb will do for now.

randomdude999:
Turns out that on windows, the patchparams struct was initialized to NULLs, which meant that 0 additional defines were being parsed. On linux it's garbage data which meant that an undefined number of additional defines (probably not 0) would be parsed.

RPGHacker:
Ah, yeah, good old missing initialization, very common error. We should probably add a memset() to test.cpp to prevent that from happening in the future.

As for Linux Debugging, I personally use CodeLite, which I like a lot so far, because it's very similar to Visual Studio. If you plan to use it, make sure to pass the correct command line parameters to cmake, though, otherwise breakpoints etc. won't work (don't know why, even when you specifically enable debug symbols, breakpoints just won't be hit). This is the command line I used:

`cmake -G"CodeLite - Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ../src`

Where the -DCMAKE_BUILD_TYPE=Debug is the important part.

RPGHacker:
Anyways, should I merge now, or do you first want to add the new parameters to the python wrapper? (Just leave a comment here whenever you think I should do the merge).

randomdude999:
All right, it builds fine and the standard defines work (didn't test standard includes, but the python wrapper code for them is identical) - i think we're good to go.

PR #44, opened 2018-03-30
Update tests

### LINKED ISSUE #39: Make test suite not require a clean SMW rom
Currently running the built-in test suite requires having a clean unheadered SMW rom. This means that running the test suite in continous integration is AFAIK impossible to do legally (without sharing a ROM somewhere so the CI can access it). Currently only 11 tests need the clean SMW rom (`grep '^;@\+' tests/*.asm` too see which ones, mostly the ones dealing with freespace). We could replace this with a dummy ROM that is the same size as SMW but is completely garbage (it should include a few large blocks of nulls/0xFF's just to make sure the freespace finder doesn't find freespace in the original ROM space). Then we could include the dummy rom in the repo somewhere.

randomdude999:
I added a dummy rom and removed all checksum checks from tests that used the SMW base. The tests still work with the SMW rom, but you can use dummy_rom.sfc aswell.
Closes #39.
Also I made Appveyor and Travis run the test suite too (currently only using the DLL test since I felt running it twice is kinda unnecessary, but tell me if i should change it. Remember that currently the app test suite fails since it treats the warning from xkasemu.asm as an error).


randomdude999:
that last one was surprisingly successful... now just need to fix that one test

randomdude999:
Hey look, appveyor now runs the built-in test suite correctly! (travis build is still broken though)

RPGHacker:
Would it be possible to update the tests to also create x64 builds of the lib and exectuable (at least on Windows)? It's not really high priority and I don't think anyone will ever really use x64 versions of Asar, but building for x64 can actually help finding some logic errors in code, and it doesn't really cause any harm to support both, 32-bit and 64-bit builds.

Also, again only if it's possible and not too much trouble, another Windows build which uses Visual Studio's clang platform toolset instead of the default VS platform toolset? Since clang has very different warnings from Windows, it too can help catching some stuff that Windows might miss.

randomdude999:
Pretty sure it's possible (the x64 at least). The Appveyor documentation on clang seems very sparse, but I'll see what I can do. We could also build with clang on Linux, that's much easier AFAIK.

RPGHacker:
In Visual Studio, you can select a platform toolset for a project. Normally it's set to "v141" in VS2017, but can be set to "v141_clang_c2" to build with clang instead. clangs supports this platform toolset, for example. If AppVeyor also supports it, might not be too dificult to set it up. Just having clang on Linux also seems fine, though (since I don't expect it to act too differently on Windows and Linux).

Anyways, thanks, that would be great!

randomdude999:
All right, got clang building working (https://travis-ci.org/RPGHacker/asar/builds/360581315). I'm pretty sure Travis already builds on x64. Now to Appveyor...

randomdude999:
All right, got Appveyor to build x64 too (https://ci.appveyor.com/project/RPGHacker/asar/build/23). I don't think I'm going to start messing with clang on windows right now, since Travis builds with clang already.

RPGHacker:
Thanks!
Yeah, having it on just one platform is already helpful enough, I think.

PR #47, opened 2018-03-31
Build with clang and gcc on linux and x64 and x86 on windows

randomdude999:
I made Travis build with both clang and gcc. Also appveyor now builds both 32-bit and 64-bit.
(also when merging this you can squash the commits)
(also turns out the cmake version provided by Travis is recent enough (3.9.2), so I removed the code that downloaded a newer version)
Also, with relatively little effort I could also build Debug builds on both windows and linux, but I don't think that would really be useful.
Also, I can't understand why we need to use GCC 6.something: 4.8 is supposed to be [c++11 feature complete](https://gcc.gnu.org/projects/cxx-status.html), yet it still throws tons of errors when building.


RPGHacker:
I assume the problem is with strdup() in libstr.h

`#ifndef _MSC_VER//I thought this one was FreeBSD-only, and now it suddenly exists in MSVC but not GCC?`

Maybe they added a strdup() function to GCC at some point, which makes it conflict with this definition here?

randomdude999:
Yeah, probably. Now I'm trying to figure out when it was added. I think the next commit might work though...

RPGHacker:
Isn't it somehow possible to force cmake to use gcc 4.8 on a project? Might be easier to find and fix the remaining build problems that way than to hope for the build on the build server to complete.

randomdude999:
Thing is I don't have any machines around that have GCC 4.8 installed. Anyways, as you can see the build is now fixed. I'll clean up the config files too.

RPGHacker:
I see, I'm personally just using a VM for that (though I'm not too familiar with Linux, so I'm not sure how one would configure everything).

Anyways, glad to see it's working now!

PR #63, opened 2018-04-04
Memory file support

Horrowind:
Added initial support for memory files. I could not figure out a good way to implement tests for this, therefore this code is currently untested.

Have you an idea how to extend the test suite to include memory files? I could add this to this pull request, but I have no clue how we want to approach this now (to be honest, I haven't really looked into the test system yet).

Also the name of corresponding the DLL function could be changed if you think it is too verbose.

### DIFF
--- a/src/asar/interface-lib.cpp
+++ b/src/asar/interface-lib.cpp
@@ -185,9 +185,7 @@ void asar_patch_begin(char * romdata_, int buflen, int * romlen_, bool should_re
 		memcpy((char*)romdata_r/*we just allocated this, it's safe to violate its const*/, romdata_, (size_t)*romlen_);
 	}
 	else romdata_r = (unsigned char*)romdata_;
-	romdata = (unsigned char*)malloc(maxromsize);
-	// RPG Hacker: Without this memset, freespace commands can (and probably will) fail.
-	memset((void*)romdata, 0, maxromsize);
+	romdata = (unsigned char*)calloc(1, maxromsize);
 	memcpy((unsigned char*)romdata, romdata_, (size_t)*romlen_);
 	if (should_reset)
 		resetdllstuff();
@@ -480,4 +478,9 @@ EXPORT mapper_t asar_getmapper()
 	return mapper;
 }
 
-
+EXPORT void asar_add_virtual_memory_file(const char* path, void* buffer, size_t length)
+{
+	if(filesystem) {
+		filesystem->add_memory_file(path, buffer, length);
+	}
+}
--- a/src/asar/virtualfile.cpp
+++ b/src/asar/virtualfile.cpp
@@ -20,7 +20,7 @@ public:
 class memory_file : public virtual_file
 {
 public:
-    memory_file(void* data, size_t length)
+	memory_file(void* data, size_t length)
 		: m_data(data), m_length(length)
 	{		
 	}
@@ -30,29 +30,29 @@ public:
 		close();
 	}
 
-    virtual void close()
-    {
-    }
+	virtual void close()
+	{
+	}
 
-    virtual size_t read(void* out_buffer, size_t pos, size_t num_bytes)
-    {
-        if(pos > m_length) return 0;
+	virtual size_t read(void* out_buffer, size_t pos, size_t num_bytes)
+	{
+		if(pos > m_length) return 0;
 
-        int diff = (int)(pos + num_bytes) - (int)m_length;
-        num_bytes -= diff < 0 ? 0 : diff;
+		int diff = (int)(pos + num_bytes) - (int)m_length;
+		num_bytes -= diff < 0 ? 0 : diff;
 
-        memcpy(out_buffer, (char*)m_data + pos, num_bytes);
-        return num_bytes;
-    }
+		memcpy(out_buffer, (char*)m_data + pos, num_bytes);
+		return num_bytes;
+	}
 
-    virtual size_t get_size()
-    {
-        return m_length;
-    }
+	virtual size_t get_size()
+	{
+		return m_length;
+	}
 
 private:
-    void*  m_data;
-    size_t m_length;
+	void*  m_data;
+	size_t m_length;
 };
 
 class physical_file : public virtual_file
@@ -174,7 +174,7 @@ void virtual_filesystem::initialize(const char** include_paths, size_t num_inclu
 	}
 
 	m_last_error = vfe_none;
-    m_memory_files.reset();
+	m_memory_files.reset();
 }
 
 void virtual_filesystem::destroy()
@@ -215,14 +215,14 @@ virtual_file_handle virtual_filesystem::open_file(const char* path, const char* 
 
 		case vft_memory_file:
 		{
-            if(m_memory_files.exists(base_path)) {
-                memory_buffer mem_buf = m_memory_files.find(base_path);
-                memory_file* new_file = new memory_file(mem_buf.data, mem_buf.length);
-                return static_cast<virtual_file_handle>(new_file);
-            } else {
-                m_last_error = 	vfe_doesnt_exist;
+			if(m_memory_files.exists(base_path)) {
+				memory_buffer mem_buf = m_memory_files.find(base_path);
+				memory_file* new_file = new memory_file(mem_buf.data, mem_buf.length);
+				return static_cast<virtual_file_handle>(new_file);
+			} else {
+				m_last_error =	vfe_doesnt_exist;
 				return INVALID_VIRTUAL_FILE_HANDLE;
-            }
+			}
 		}
 
 		default:
@@ -274,17 +274,17 @@ size_t virtual_filesystem::get_file_size(virtual_file_handle file_handle)
 virtual_filesystem::virtual_file_type virtual_filesystem::get_file_type_from_path(const char* path)
 {
 	if(m_memory_files.exists(path)) {
-        return vft_memory_file;
-    } else {
-        return vft_physical_file;
-    }
+		return vft_memory_file;
+	} else {
+		return vft_physical_file;
+	}
 }
 
-void virtual_filesystem::create_memory_file(const char* name, void* buffer, size_t length) {
-    memory_buffer mem_buf = { buffer, length };
-    m_memory_files.remove(name);
-    m_memory_files.create(name) = mem_buf;
-    
+void virtual_filesystem::add_memory_file(const char* name, void* buffer, size_t length) {
+	memory_buffer mem_buf = { buffer, length };
+	m_memory_files.remove(name);
+	m_memory_files.create(name) = mem_buf;
+	
 }
 
 bool virtual_filesystem::is_path_absolute(const char* path)
--- a/src/asar/virtualfile.cpp
+++ b/src/asar/virtualfile.cpp
@@ -17,7 +17,43 @@ public:
 	virtual size_t get_size() = 0;
 };
 
+class memory_file : public virtual_file
+{
+public:
+    memory_file(void* data, size_t length)
+		: m_data(data), m_length(length)
+	{		
+	}
 
+	virtual ~memory_file()
+	{
+		close();
+	}
+
+    virtual void close()
+    {
+    }
+
+    virtual size_t read(void* out_buffer, size_t pos, size_t num_bytes)
+    {
+        if(pos > m_length) return 0;
+
+        int diff = (int)(pos + num_bytes) - (int)m_length;
+        num_bytes -= diff < 0 ? 0 : diff;
+
+        memcpy(out_buffer, (char*)m_data + pos, num_bytes);
+        return num_bytes;
+    }
+
+    virtual size_t get_size()
+    {
+        return m_length;
+    }
+
+private:
+    void*  m_data;
+    size_t m_length;
+};
 
 class physical_file : public virtual_file
 {
@@ -138,6 +174,7 @@ void virtual_filesystem::initialize(const char** include_paths, size_t num_inclu
 	}
 
 	m_last_error = vfe_none;
+    m_memory_files.reset();
 }
 
 void virtual_filesystem::destroy()
@@ -178,9 +215,14 @@ virtual_file_handle virtual_filesystem::open_file(const char* path, const char* 
 
 		case vft_memory_file:
 		{
-			// TODO: Add support for in-memory files here later
-			m_last_error = vfe_unknown;
-			return INVALID_VIRTUAL_FILE_HANDLE;
+            if(m_memory_files.exists(base_path)) {
+                memory_buffer mem_buf = m_memory_files.find(base_path);
+                memory_file* new_file = new memory_file(mem_buf.data, mem_buf.length);
+                return static_cast<virtual_file_handle>(new_file);
+            } else {
+                m_last_error = 	vfe_doesnt_exist;
+				return INVALID_VIRTUAL_FILE_HANDLE;
+            }
 		}
 
 		default:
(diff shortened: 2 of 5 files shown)


RPGHacker:
This implementation actually seems a lot smarter than what I had originally planned (my original plan was to encode the addresses to memory files directly into paths, but adding memory files via the DLL API is actually way simpler and better).

There doesn't seem to be a reset for memory files right now, or is there? Since they're currently set via a function, you'd need to reset them somewhere, presumeably in resetdllstuff().
Personally, I would just work them into patchparams for asar_patch_ex() instead. Of course that way they couldn't be used with asar_patch(), but I don't really think they're needed in in there, anyways. I consider asar_patch() a legacy function that exists only for backwards-compatibility so that existing tools can upgrade to new Asar versions without code changes, and of course without code changes, they're not going to make use of memory files anyways, yet if they're going to change the code, they might as well go ahead and replace asar_patch() by asar_patch_ex(), that's why I'd generally go with asar_patch_ex() when adding any new parameters. Let me know if you disagree here and if you see good reasons to have this functionality as a separate function, anyways.

Regarding the implementation of memory files, there's one (potential) problem that might arise and needs to be considered: the file paths that are passed to the API need to match whatever paths Asar uses to open them, otherwise it won't find the memory files and opens the files regularly instead. Basically, there's a couple of situations where I think this can currently (and probably will) go wrong:
* When the API is passed absolute paths and a patch references a relative path (or vice versa)
* When the path separators in the API and in a patch don't match exactly (for example: API using backward slashes and patch using forward slashes)
* Even when both API and patch use a matching relative path, stuff can still go wrong. Imagine adding a file includes/header.asm via the API, then assembling a patch1.asm which does `incsrc includes/header.asm`, but also includes another patch patch2.asm via `incsrc patch2/patch2.asm`, which itself does `incsrc includes/header.asm`, we now have two different ASM files referencing two different includes that are both accessed via the same path, so Asar would probably pick the wrong file from memory here.

Fixing all of these problems seems a bit tough, but I have a few ideas for this:
* Have both asar_patch_ex() and asar_add_virtual_memory_file() require explicitly providing absolute paths only (for all parameters that take a file path). When a relative path is used anywhere, make the function fail with an error message. That way, we can prevent our working directory from factoring into anything. Since both APIs are new and thus can't break any compatibility yet, adding this requirement should be acceptable. Actually it's yet another reason why I would disregard asar_patch() for memory files: we can't enforce absolute paths in asar_patch() without potentially breaking compatibility, but without enforcing absolute paths, solving this problem would be even more difficult (factoring the working directory into stuff gets really, really nasty).
* Whereever paths are created in Asar, we need to remove/resolve all ../ and ./ path components so that we don't accidentally end up with two different-looking paths that link to the same file (includes/blah/../header.asm is the same as includes/header.asm, but a string compare will only work after resolving the first path).
* Whereever paths are created in Asar, we need to convert all backslahes to forward slashes (there's really no reason to even use backslashes on Windows these days, I don't think any relevant I/O API doesn't support forward slashes).
* We need the include search path walker from physical_file on a higher level, that is, directly in the virtual filesystem (if I haven't moved it there already, I'm not sure). Since we index memory files via their paths, we of course need to know the final path to a file so that we can correctly find it in memory, and this, of course, only works when we apply include search paths before picking the actual interface. Basically, we should take the code from physical_file's open_file(), and then whereever it checks file_exists(), we do a file_exists_in_memory() first (to prioritize files in memory, while still maintaining include search path priority).

Hope I didn't forget anything. It's certainly way complexer than I had originally estimated. In fact, most of these things just crossed my mind while writing this comment, but I think they're necessary to consider to make the features work well.

As for how to test them, I persoanlly recommend working them into the DLL test suite. Currently, it creates temporary files from the source files, writes them to disk and then passes their path to the DLL. We could skip most of this and just pass the temporary files to asar_patch_ex() directly as memory files. Of course the app test suite would still have to use the old way of writing out temporary files (oh well, technically, it doesn't even have to, it can just use the original source files, it will just make it throw thousands of warnings).

Longest comment I've probably written on GitHub so far, but TL;DR: this is more complex than I had originally anticipated. :V

Horrowind:
I also thought about this for some time.

For some reason I dislike being forced to have absolute paths for memory files. If we want to go this way, we could actually just prefix them similar to "mem://path/to/file" to distinguish between the two. The reason I dislike them is because it makes sense to also write memory files to disk if I am done with editing with the tool, so that a user can just use Asar with the .asm files without running the tool. With this I could change sprite ASM for example and run Asar instead of sprite tool, which I only need to do if I changed the configuration of a sprite.


Thus I think that this has to work like this:
1. On creation of the virtual file system, we provide a list of include directories, which works like the working directory of the virtual file system. Let's just call them include directories from now on. Every memory file is relative to every include directory if it is a relative path (or it is absolute and uses the usual root). We could also distinguish between the working directory, and the include directories. Then we would only handle paths which are relative to this working directory. In the following I assume handle memory files relative to all include directories, but the semantics easily copy over, if we only have a single working directory.

I don't think that multi-threading or use of a shared library is a problem here. We would pass the base name of the file as one working directory and the usual include directories or let the user of the DLL pass something to us which we would use.

2. Every path of memory file I give to the virtual file system will be normailzed. This means that we will prepend every working directory to all relative paths and that we kill all instances of "./" and "../". If we want to support symlinks we also need to walk along this normalized path until we get stuck and replace all symbolic links with their target. Oh, and \ to / of course. We add these normalized file names to our hash map. 

Example: Suppose we have the include directories `/just/some/path/` and `/path/to/a/symlink/` where `symlink` resolves to `/some/target/`, and want to add a memory file at `./../bla/test.asm`. We would add the following two items to our hash map: `/just/some/bla/test.asm` and `/some/bla/test.asm`.

If we want to ignore symlinks, then we would only need to kill "./" and "../", I think.

3. If I want to look up a file given by some path we do the same procedure. If we find the normalized path in our hash map - great, otherwise open the files (one for each include directory) regularly. If all fail this is an error. This has a small error: Usually `/some/invalid/path/../../../some/regular/file.asm` gives an error, when `/some/invalid/path` is not found, but I have seen some tools which open these fine and I would not worry to much about it.

Does this makes sense to you? A small note on your example: The first file should include `includes/header.asm` while the second file should include `patch2/includes/header.asm`. But this needs to be handled before the path is given to the virtual file system. But maybe I am ignoring how Asar has handled relative paths of included patches in the past and we need to be compatible with previous versions.

The use of functions versus large parameter structs is something I think people call immediate mode versus retained mode API design. The profit of the function call is that I can provide the API my data the way I already have it. In the parameter struct I need to provide an array, which I need to allocate, then I need to store my string pointers and then after the call I need to deallocate this array. In the case of function calls I can just iterate over all file names like I would normally do and don't need to convert them into some new structure. So the whole difference here is just the repackaging, which we could either do on our side or we require our users that they do it.

Implementation wise it is easier to use a parameter struct, since we can only do step two when we know all include directories. Thus we would need to store every relative path the user gave us somewhere and would do normalization only when `asar_path_ex` is called.

On the whole using a more immediate mode API would only make sense if we would do this consistently. But since this is Asar, I fear that consistency is simply not a design goal :/.

Considering our options, I would vote for supporting symlinks and having virtual files only relative to the working directory and not to the include directories, but I am not so sure about the latter. Let me know what you think about this approach. If you give your okay and I have time to work on it, I might.

Horrowind:
I am just glimpsing over it right now. Could it be, that your running example is wrong? I find it odd that you want to include `my_game/sub/a.asm` four times. Is it just `my_game/a.asm` in 2. and 3. ?

Just tell me if I am wrong here, you don't need to write another wall of text. I will read through it more careful then.

RPGHacker:
It looks for it four times, because `my_game/sub` is an include search path, yet it is also the path where `b.asm` resides. The include finder first checks the latter (as in, a path relative to the file currently being assembled), and if it doesn't find the file there, it looks in the include search path. In this example, the include search path just so happens to result in the same path, but of course that's not always the case. And the "four times" (as opposed to just "two times") of course come from the fact that I'm already accounting for "in-memory" files here (we should always check "in-memory" first so that we don't read files from disk that already exist in memory). Of course without in-memory files, we would only have to check two times, which is what the current build does.

Let me pick a different include path here, so that the example becomes more obvious. Let's say we had an include path "C:/libsmw". The current Asar version would check the following paths in this order:

1. `my_game/sub/a.asm` on disk
2. `C:/libsmw/a.asm` on disk
3. File not found, fail

now if we accounted for in-memory files, this should change to

1. `my_game/sub/a.asm` in memory
2. `my_game/sub/a.asm` on disk
3. `C:/libsmw/a.asm` in memory
4. `C:/libsmw/a.asm` on disk
5. File not found, fail

So yeah, the fact that it's checking the same path four times in my example above just happened by chance, of course it wouldn't always do it like that.

Also nice weather today, isn't it? :V

Horrowind:
I will split things up into multiple parts here:
Usage of relative paths in general: The thing is that requiring absolute paths in the API makes things complicated for the library user, since he now has to handle to paths instead of one. The first one is the absolute path he would need to put in the patchparams struct. The second path is the one used in the patch where he writes `incsrc relative/path/to/patch`. Of course I can also write an absolute path into the patch, but if I want to save them to disk they are now position dependent and I can not move them in my file system.

And since we are struggling to find the right semantics for relative paths here, we should not force this struggle also on the user of the library. Because if he now needs to think about how Asar handles relative paths and needs to do all the steps we are arguing about here.  In the end this would be like: "Nah, we don't really know how to do this, but you sure can figure it out!"

Horrowind:
**The algorithm part:**

I think you are saying that what I wanted to merge into a single concept, include directory and the working directory needs to be actually two concepts. (I had in mind, that the working directory is conceptually just one of the include directories.) 

This is fine with me. Include directories are those given in the patchparams struct, while the  working directory is the base directory of the file name of the patch. So the procedure would be that step 2 of my procedure would simplify to just prepending the working directory before normalizing. (I meant to write include directories instead of working directories in the beginning of step 2 above).

In step 3 we would still need to prepend the working directory and every include directory to a relative path. This is for the case where I added `includes/` to the include directories and `includes/a.asm` as a memory file and now want to incsrc `a.asm`, which I expect to work.

This should recreate both of the examples in your last and second to last post.

Horrowind:
> What exactly do you mean by "before" here, or rather, when is this "before"? The earliest possible time at which Asar could determine the path to an included file is when the actual include command (incsrc, incbin etc.) is encountered. Are you proposing that the filesystem should know the paths of includes before that? I don't understand how that would work.
Or are you proposing that patch2.asm should do incsrc patch2/includes/header.asm instead of just includes/header.asm? That seems rather weird. Asar has always treated include files as being relative to the currently assembled file. This also seems way more intuitive, and it makes it possible to drag around whole patch directory structures without entire include hierarchies breaking (you only need to change includes in the files that didn't move as opposed to having to change includes in all files).

Ahh, okay, I think I understand now. What I missed is that we need to assume that the working directory is dynamic. So this adds a forth step to the algorithm:
4. If we encounter `incrc "some/path"`, we resolve `some/path` as in step 3, and set the working directory to the base directory of `some/path`.

This should now solve every issue I have.

RPGHacker:
> The thing is that requiring absolute paths in the API makes things complicated for the library user

True, that's why I wouldn't make it a hard requirement and only warn. Technically, there should be very few situations were relative paths could break something (the only case I can spontanously think of is when someone uses absolute paths in an include, while using a relative path in the API). Still the user should at least now that relative paths can potentially break something. Technically, it's possible to get absolute paths from relative paths on both Windows and Linux, I think, but from what I know, it requires the paths to actually exist (which of course doesn't necessarily have to be the case for memory files, depending on how you use them).

> since he now has to handle to paths instead of one. The first one is the absolute path he would need to put in the patchparams struct. The second path is the one used in the patch where he writes incsrc relative/path/to/patch.

Actually, not quite, I think you're completely misunderstanding what my actual proposal is, which might also be why this discussion is taking place in the first place. Include paths in patches are completely unaffected, they can stay relative, because when the current Asar assembles a file, it always internally turns paths to includes into something like "parent file path + include file path" (or "include search path + include file path" if that's where it found the file). What this means is that only the API needs to get a absolute paths (for the base file and all include search paths), because if it does, all subsequent include paths automatically become absolute paths as well. Of course if the parent paths/include search paths are relative paths, all paths internally to Asar will also be relative paths. Since only the API input should be absolute paths, the burden for the tool coder/tool user should be relatively little (to non-existent if we only warn on relative paths).

> I had in mind, that the working directory is conceptually just one of the include directories.

Yeah, I've thought about this as well in the past. In fact, my original idea with include finding was to just set the path to the current file as the first include path. Decided against it because if we did that, we would have to overwrite the first include path every time we included an asm file (and thus the path to the current file changed), that's why I'm checking the current file specifically. But yeah, conceptually, they're pretty much the same.

> So the procedure would be that step 2 of my procedure would simplify to just prepending the working directory before normalizing.

That would be a possibility, of course. I mean in the end, it would come down to the same as letting the tool coder prepend the working directory to every single include search path and the base path before passing it to the API. My approach needs a bit more "effort" from the coder, while your approach needs an addtional parameter. Of course if we use my approach and make it optional (with only warns), no extra effort is needed, unless they want to be on the save side.

(Help, you're too quick for me - will probably reply to your last post when I'm back home)

randomdude999:
Slightly unrelated to this discussion but Horrowind please fix your git setup: currently the commit history will show like 3 commits by someone called "Your Name", and you aren't listed under the contributors tab of the insights page. Try `git config --global user.name Horrowind` and `git config --global user.email email.you.registered.to.github.with@example.com`.

Horrowind:
Yeah I noticed that too, thanks for reminding me.

RPGHacker:
I've just merged support for includeonce, together with my version of normalize_path(). I went with a slightly different solution than what you proposed in the UTF-8 issue. I tried your version at first, but it failed in a couple of simple scenarios I threw at it (including the xkasemu.asm test), so I went ahead and wrote my own version (actually tried two or three different versions to end up with one I'm satisfied with). My current solution still isn't perfect. Invalid paths will likely break it, but I don't care about that since they're, after all, invalid. It also likely won't work with Windows network paths (prefixed with \\ ), but I hope noone uses Asar that way, anyways.

I also didn't include the working directory parameter in my version of the function. Currently, the lib interface just warns on relative paths (I think the standalone executable does nothing). If we're going to add the working directory parameter, I feel it should be optional and we shouldn't pass it to this function, but rather already prepend it to all relative paths as they're passed into the library or application (just my gut feeling, not sure if actually the better solution).

Also didn't add support for symbolic links after all. On Windows, I think that features is Vista or later only, and generally speaking it also requires that a files exists. Of course we could differentiate between physical files and memory files by calling file_exists(), but I don't think it's too important, so right my function is pure string manipulation.

Anyways, just a little update. Feel free to use my function in your memory file API. I think all you would have to do would be to call the function in your setter function (or wherever the paths are passed in).

RPGHacker:
I guess that's one good reason to go with my proposal after all.

src/asar/virtualfile.cpp
```diff
@@ -178,9 +215,14 @@ virtual_file_handle virtual_filesystem::open_file(const char* path, const char*
 
 		case vft_memory_file:
 		{
-			// TODO: Add support for in-memory files here later
-			m_last_error = vfe_unknown;
-			return INVALID_VIRTUAL_FILE_HANDLE;
+			if(m_memory_files.exists(base_path)) {
```
randomdude999: Are you *sure* that this is supposed to check `base_path`?

src/asar/interface-lib.cpp
```diff
@@ -480,4 +478,9 @@ EXPORT mapper_t asar_getmapper()
 	return mapper;
 }
 
-
+EXPORT void asar_add_virtual_memory_file(const char* path, void* buffer, size_t length)
```
randomdude999: Already mentioned this but this function doesn't do anything since the filesystem is only created in `asar_patch`(`_ex`).

src/asar/virtualfile.hpp
```diff
@@ -21,6 +22,12 @@ enum virtual_file_error
 	vfe_num_errors
 };
 
+struct memory_buffer
+{
+    void*  data;
```
randomdude999: Is there any reason this couldn't be `const`?

randomdude999:
I have merged this on the memory-file-support branch. I have also fixed some things about it (the 3 issues i mentioned, plus removed a warning when compiling on clang). I feel that it's good to merge now.

RPGHacker:
Went ahead and created a new pull request based on your changes. It can be found here:
https://github.com/RPGHacker/asar/pull/101
Also going ahead and closing this one. If we need to discuss stuff further, we can do so on the new pull request.

PR #64, opened 2018-04-06
Fix some SA-1 issues

### LINKED ISSUE #61: SA-1 mapping broken for some addresses
TheBiob on discord (in a DM with me):
> also, telinc reminded me of another bug I forgot yesterday, If I'm correct somethings broken with the pctosnes routine for sa-1 roms
> If I'm correct then my debugging a while ago showed that the 0x0F8000 can cause it to mask out the high nibble of the bank byte causing it to overwrite a completely different bank to what it's supposed to be

("the 0x0F8000" probably references libsmw.h line 157)

Not sure if this is still relevant and even if it is then how to fix it, but I think it's worth a look.

randomdude999:
Closes #61.

### DIFF
--- a/src/asar/libsmw.h
+++ b/src/asar/libsmw.h
@@ -154,7 +154,7 @@ inline int pctosnes(int addr)
 	{
 		for (int i=0;i<8;i++)
 		{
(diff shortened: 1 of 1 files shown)


randomdude999:
Telinc1 just tested and said it broke his rom even harder than with asar 1.50.

PR #77, opened 2018-04-14
Update changelog

randomdude999:
I added information about who did what to the changelog, because why not?

### DIFF
--- a/docs/changelog/index.html
+++ b/docs/changelog/index.html
@@ -56,8 +56,7 @@
 
 		<h3>New features:</h3>
 		<ul>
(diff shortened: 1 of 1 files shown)


randomdude999:
When merging future pull requests, could you please still do them with a merge commit instead of squashing? I suggested squashing that one pull request because each individual commit there had very little value and would clutter up the commit history, but squashing messes up the network graph and makes git think that the branch isn't fully merged (thus throwing an error when trying to delete the branch on my computer). Also it makes the statistics under the insights tab less accurate because multiple commits are shown as one.

PR #78, opened 2018-04-15
Using strcpy is defective in string::assign.

Horrowind:
Today I tried to find bugs in my code using `valgrind`. To be sure that I am the author of these bugs, I compared them with the current `master` branch, and this is the only issue I found. It's not a huge issue, but I might as well fix it. `strcpy` requires src and dest to be non-overlapping (http://en.cppreference.com/w/cpp/string/byte/strcpy).

This is the case in `interface-cli.cpp:250`,`interface-cli.cpp:251`,`interface-cli.cpp:267`, and `interface-cli.cpp:268`.

### DIFF
--- a/src/asar/libstr.h
+++ b/src/asar/libstr.h
@@ -46,7 +46,11 @@ void assign(const char * newstr)
 	if (!newstr) newstr="";
 	len=(int)strlen(newstr);
 	resize(len);
-	strcpy(str, newstr);
+    for(int i = 0; i < len; ++i)
+    {
+        str[i] = newstr[i];
+    }
+    str[len] = '\0';
 }
 
 void assign(const char * newstr, int newlen)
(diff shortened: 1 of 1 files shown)


randomdude999:
You've made your git setup better, but the commits still aren't assigned to your github account - are you sure that the email you set with `git config --global` is verified with your github account?

Also, why did you make sure to only look for bugs made by you? You should probably search for bugs by other people too.

Horrowind:
I have hidden my email address on github, so I don't know how that could work, if I want to leave it that way.

Soon after pushing this I realized that this is still wrong and that I should use `memmove` instead. New pull request incoming...

randomdude999:
It should work fine with hidden emails - just make sure that the email you set up git with is in your verified emails list.

Issue #34, opened 2018-03-18
Add new asar_patch_ex() API function to the other bindings


RPGHacker:
Recently added an asar_patch_ex() function to the DLL C API, but forgot to add it to the other APIs (like C#). There are tools out there which use the C# binding in particular, so it's something that should be added before the next release.

randomdude999:
I added it to the python wrapper (although I'm not too sure the include path thing is correct), but I don't know enough C# to update that wrapper.

RPGHacker:
I can probably do the C# wrapper myself, I think. Just haven't really gotten to do anything Asar-related those past few days.

randomdude999:
For the python wrapper it would be more convenient to just always use asar_patch_ex instead of asar_patch. Would that be okay?

RPGHacker:
Yep, that should be no problem. Since nothing uses the Python wrapper yet, there isn't really any compatibility that can break.

randomdude999:
Closed by #80.

PR #80, opened 2018-04-22
Add Symbols file generation support

KevinCathcart:
I've added the ability to create symbols files in two different formats, corresponding to the two SNES debuggers that support symbols files that I know about. WLA format can be 
used by Benjamin Schulte's [improved BSNES-plus debugger](https://github.com/BenjaminSchulte/bsnes-plus), and nocash format can be used by no$sns.

The Library API simply returns the contents of the symbol file as a string, allowing the host application to control the actual creation of the file.

I'm not particularly satisfied with the implementation details, and I'm not sure if the library API is reasonable, so I've not yet bothered to update the wrappers, but everything does seem to function.


randomdude999:
what the fuck that debugger looks awesome, why have i not found it earlier

anyways, this looks nice, but would you mind adding the new function to the DLL wrappers (asardll.h, maybe asar.py and asar.cs too if you know them)?

KevinCathcart:
Yeah, definitely not super safe. Both debuggers do accept the symbol file with the current naming pattern. Let me see how they like the .sfc.sym pattern. If both like it, than great. I certainly have no objecting to adding a "--symbol-out <filename>" option as well. 

And yeah, I'll add it to the wrappers. I just wanted RPGHacker to have a chance to comment on the api before i bothered doing that.

RPGHacker:
Think the API is fine, can't really think of a better way to handle it with the current code, so feel free to add the missing wrappers if you can.

randomdude999:
Also good job, you broke the osx build. It's probably because `basename` is a library function too.

KevinCathcart:
Yeah. Copy an paste error in the python one. I'll fix and repush. As for the broken build, I'm trying to figure out how bsnes-plus does not have the same problem despite nall defining basename, and will replicate their solution.

randomdude999:
It is either because of the `inline`, or because nall doesn't include `string.h` (which I doubt, but might also be possible).

RPGHacker:
The simplest (and safest) is to just rename the function. I would probably go with something like "get_file_basename()" or similar.

KevinCathcart:
Yeah I renamed it to get_base_name, good enough. bsnes gets away with it due to actually using c++ namespaces.

KevinCathcart:
At this point everything mentioned so far has been addressed. I've updated the bindings (including bringing the .net binding up to date because it was lagging behind). I've personally tested the .Net binding and it works. I've not directly tested the python or c bindings.

RPGHacker:
One of the many reasons why namespaces are useful. Would definitely use them in an Asar rewrite if one was to happen.

Thanks! Will make sure to merge it later (might take a bit, since I'm kinda finding myself a busy phase again right now).

randomdude999:
In that case, this PR will also close #34.
(also overwriting git history :angry:)

KevinCathcart:
Hmm. On second thought, I technically broke ABI compatiability with the .net binding (for the case where somebody compiles it to its own dll), while retaining source level compatibility. If we care about that I can fix it easily enough. But it has not been officially released as a compiled DLL, so it may not matter.

RPGHacker:
Reminds me that it would be super practical to have tests for the bindings as well, otherweise it's really easy to forget updating them. Seems like quite some effort, though, and kinda beyond the scope of what should be focussed on.

RPGHacker:
Yeah, I don't think anyone uses the CS binding as a DLL. Don't really see a point in that, anyways, since you would be using a DLL using a DLL, which seems kinda unnecessary, so I think it should be fine the way it currently is.

src/asar-dll-bindings/asar.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, dllname):
             self.setup_func("getwrittenblocks", (c_int_ptr,),
                             POINTER(writtenblockdata))
             self.setup_func("getmapper", (), c_int)
+            self.setup_func("getsymbolsfile", (c_char_p), c_char_p)
```
randomdude999: You missed a comma here, it's supposed to be `(c_char_p,)` not `(c_char_p)`. Otherwise, Python automatically removes the parentheses instead of creating a tuple with 1 element.

src/asar-dll-bindings/asar.py
```diff
@@ -350,3 +351,11 @@ def getwrittenblocks():
 def getmapper():
     """Get the ROM mapper currently used by Asar."""
     return mappertype(_asar.dll.asar_getmapper())
+
+def getsymbolsfile(format):
```
randomdude999: This overrides a builtin (not a very big issue, but I'd rename it to `fmt` for example). Also, I would supply a default here to match the default of the CLI. (so `fmt="wla"` instead of `format`).

src/asar-dll-bindings/asar.py
```diff
@@ -350,3 +351,11 @@ def getwrittenblocks():
...
+    """Generates the contents of a symbols file for in a specific format.
+
+    Returns the textual contents of the symbols file.
+    format specified the format of the symbols file that gets generated.
+    """
+    return _asar.dll.asar_getsymbolsfile(format)
```
randomdude999: I think this should use strings, not bytes. Put `encode` and `decode` here, like on line 311.

src/asar-dll-bindings/asar.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, dllname):
             self.setup_func("getwrittenblocks", (c_int_ptr,),
                             POINTER(writtenblockdata))
             self.setup_func("getmapper", (), c_int)
-            self.setup_func("getsymbolsfile", (c_char_p), c_char_p)
+            self.setup_func("getmapper", (c_char_p,), c_char_p)
```
randomdude999: Now you changed this one back to `getmapper` instead of `getsymbolsfile`...

KevinCathcart: WTF. My editor must be acting screwy for that to happen.

KevinCathcart:
Hopefully that is better now.

RPGHacker:
Just a typical day in a programmer's life. 🙂

Issue #13, opened 2018-03-07
Make warnings more customizable


RPGHacker:
Make warnings more customizable like in MSVC, clang etc.

Some examples:
-Add support for "warnings as errors"
-Add support for warnings levels (print only warnigns of the current level or higher)
-Give all warnings an ID/number and add support for disabling/enable warnings per its ID
-Maybe support for disabling/enabling certain warnings or pushing/popping warnings settings from within an asm file (for example: @pushwarnings and @pullwarnings)

randomdude999:
List of warnings generated by Asar
```
manual pass check:
assembleblock.cpp:1030: Manual warn without argument (pass 2)
assembleblock.cpp:1038: Manual warn with argument (pass 2)
assembleblock.cpp:1054: \ Expecttitle mismatch, title checks disabled
assembleblock.cpp:1079: / (pass 0)

warn0:
arch-65816.cpp:137: ($yy),x does not exist, assuming $yy,x
arch-65816.cpp:157: $xx,y is not valid with 8-bit parameters, assuming 16-bit
arch-spc700.cpp:318: This opcode does not exist with 16-bit parameters, assuming 8-bit
assembleblock.cpp:60: missing org or freespace command
assembleblock.cpp:359:  \
assembleblock.cpp:854:  |
assembleblock.cpp:1349: | Asar-only commands in ;@xkas patch
assembleblock.cpp:1525: |
assembleblock.cpp:1556: |
assembleblock.cpp:1962: |
main.cpp:372:           /
assembleblock.cpp:1129: \ xkas deprecation warning
assembleblock.cpp:1990: /
assembleblock.cpp:1158: db "STAR" warning
assembleblock.cpp:1224: org $005000 in LoROM
assembleblock.cpp:1631: warnpc is relaxed 1 byte
assembleblock.cpp:1647: rep -1 warning
assembleblock.cpp:1662: \ xkas incsrc relative to assembler warning
assembleblock.cpp:1701: /
assembleblock.cpp:1674: \ '\' in file path warning
assembleblock.cpp:1713: /
assembleblock.cpp:2010: Unrecognized special command
main.cpp:354: parentheses after defines

warn1:
main.cpp:134: label optimization warning

warn2:
assembleblock.cpp:1419: \ Freespace leak warning
assembleblock.cpp:1755: /
```
Now we should probably assign each one an ID and possibly a short (1 or 2 hyphen separated words) description.

RPGHacker:
Nice work!
[...]

PR #81, opened 2018-04-22
New optional warning and bugfix

KevinCathcart:
It should be possible to completely avoid having asar silently create non-working code due to assumptions. Using `bank noassume` brings us most of the way there. But immediate mode instructions can still be incorrectly assembled if the user fails to specify `.b` or `.w`. Asar's heuristics are pretty good and usually get things right, especially if the value is a hexadecimal constant. However, it can guess wrong on occasion, especially in cases involving calculations.

To allow detecting these I've added a new optional warning that can be enabled with "warn immed on", that warns if an immediate instruction has implicit size and is therefore relying on the heuristics for correctness. If size is explicitly specified with `.b` or .`w`, the warning is obviously not generated.

This also let me fix the following (from the 1.5.0 manual.txt's Known Bugs section):
>JML.w and JSL.w aren't rejected; they're treated as JML.l and JSL.l. This is since Asar has a much
  more powerful size finding system than xkas, and automatically picks .w if it contains a label in
  the same bank - even for .l-only instructions like JML. Therefore, I'm forced to either reject JML
  and JSL to a label in the same bank, or allow JML.w, and I prefer allowing unusual but bad stuff
  over rejecting common and valid stuff. Workaround: Don't do this.

This was fixed by emit an error only if a size of .w is explicitly specified, while permitting a calculated size of .w to be used without error.

### DIFF
--- a/src/asar/arch-65816.cpp
+++ b/src/asar/arch-65816.cpp
@@ -32,9 +32,10 @@ bool asblock_65816(char** word, int numwords)
 	unsigned int num;
 	int len=0;//declared here for A->generic fallback
 	bool explicitlen = false;
+	bool hexconstant = false;
 	if(0);
 	else if (assemblemapper(word, numwords)) {}
-#define getvars(optbank) num=(pass!=0)?getnum(par):0; if (word[0][3]=='.') { len=getlenfromchar(word[0][4]); explicitlen=true; word[0][3]='\0'; } else {len=getlen(par, optbank); explicitlen=false;}
+#define getvars(optbank) num=(pass!=0)?getnum(par):0; hexconstant=is_hex_constant(par); if (word[0][3]=='.') { len=getlenfromchar(word[0][4]); explicitlen=true; word[0][3]='\0'; } else {len=getlen(par, optbank); explicitlen=false;}
 #define match(left, right) (word[1] && stribegin(par, left) && striend(par, right))
 #define init(left, right) itrim(par, left, right); getvars(false)
 #define bankoptinit(left, right) itrim(par, left, right); getvars(true)
@@ -47,9 +48,9 @@ bool asblock_65816(char** word, int numwords)
 #define as3(    op, byte) if (is(op) && len==3) { write1((unsigned int)byte); write3(num); return true; }
 //#define as23(   op, byte) if (is(op) && (len==2 || len==3)) { write1(byte); write2(num); return true; }
 #define as32(   op, byte) if (is(op) && ((len==2 && !explicitlen) || len==3)) { write1((unsigned int)byte); write3(num); return true; }
-#define as_a(   op, byte) if (is(op)) { if(!explicitlen && warnimpimmed) warn0("implicitly sized immediate"); if (len==1) { write1(byte); write1(num); } \
+#define as_a(   op, byte) if (is(op)) { if(!explicitlen && !hexconstant && warnimpimmed) warn0("implicitly sized immediate"); if (len==1) { write1(byte); write1(num); } \
 																					 else { write1((unsigned int)byte); write2(num); } return true; }
-#define as_xy(  op, byte) if (is(op)) { if(!explicitlen && warnimpimmed) warn0("implicitly sized immediate"); if (len==1) { write1(byte); write1(num); } \
+#define as_xy(  op, byte) if (is(op)) { if(!explicitlen && !hexconstant && warnimpimmed) warn0("implicitly sized immediate"); if (len==1) { write1(byte); write1(num); } \
 																					 else {  write1((unsigned int)byte); write2(num); } return true; }
 #define as_rep( op, byte) if (is(op)) { if (pass==0) num=getnum(par); for (unsigned int i=0;i<num;i++) { write1((unsigned int)byte); } return true; }
 #define as_rel1(op, byte) if (is(op)) { int pos=(!foundlabel)?(int)num:(int)num-((snespos&0xFFFFFF)+2); write1((unsigned int)byte); write1((unsigned int)pos); \
--- a/src/asar/arch-65816.cpp
+++ b/src/asar/arch-65816.cpp
@@ -46,7 +46,7 @@ bool asblock_65816(char** word, int numwords)
 													/*if (is(op) && len==3 && emulate) { write1(byte); write2(num); return true; }*/
 #define as3(    op, byte) if (is(op) && len==3) { write1((unsigned int)byte); write3(num); return true; }
 //#define as23(   op, byte) if (is(op) && (len==2 || len==3)) { write1(byte); write2(num); return true; }
-#define as32(   op, byte) if (is(op) && (len==2 || len==3)) { write1((unsigned int)byte); write3(num); return true; }
+#define as32(   op, byte) if (is(op) && ((len==2 && !explicitlen) || len==3)) { write1((unsigned int)byte); write3(num); return true; }
 #define as_a(   op, byte) if (is(op)) { if(!explicitlen && warnimpimmed) warn0("implicitly sized immediate"); if (len==1) { write1(byte); write1(num); } \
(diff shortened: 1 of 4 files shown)


KevinCathcart:
Hmm. On second thought, I technically broke ABI compatiability with the .net binding (for the case where somebody compiles it to its own dll), while retaining source level compatibility. If we care about that I can fix it easily enough. But it has not been officially released as a compiled DLL, so it may not matter.

KevinCathcart:
I should probably clarify the intended purpose behind this.

When an assembler has to resort to heuristics it is possible that it will sometimes pick a size different from what the coder intended. These errors can fall into two categories.

The first category is often not harmful. If the assembler believes that a long address is needed to access some location when a short address is possible and intended, the resulting code will ususally still work, albeit slower. It can be a problem is some cases, especially when patching existing code in-line, and also in timing sensitive code, but those are usually a small fraction of the total code.

The other category of errors are those that are almost always a problem. If the assembler thinks it can use a short address when a long address is required, the code will malfunction. Similarly if the assembler guesses the wrong size for an immediate mode instruction, the code will likely malfunction. You can mitigate the first case by using `bank noassume`. But the second case requires that you either: 1) always use hex constants because they will always be the right size, 2) explicitly use .b or .w, or 3) get lucky with the heuristics never guessing wrong. While obviously case 3 works for many people, having an option to not rely on it would be helpful. 

Hence this patch. It enables a warning whenever you leave .b or .w off of an immediate mode instruction. I do want to extend this patch further so that it also does not warn for hex constants. The end result would be that it will only warn in case 3. 

Obviously nothing will help if the developer types the wrong suffix or uses the wrong sized hex constant, but at least in those cases the developer's code is objectively wrong, and it is not the assembler guessing wrong.

Footnote:
You might be able to get away with the assembler choosing 16 bit mode wrongly if the upper byte of the immediate is zero, and you have a do-nothing `brk` handler, but obviously that has a rather large performance penalty!

RPGHacker:
It's definitely a useful and needed feature that will be very handy in a couple of situations, I'm sure :)
Will just take me a while to merge it, busy as I currently am.

KevinCathcart:
No problem, especially since without the planned enhancement it is not nearly as useful as right now as it should be. I just wanted to be clearer on the intended purpose, as I realized my initial wording left something to be desired. 

Posting it early was mostly to avoid having somebody else start working on a similar feature, and causing duplicated development effort.

RPGHacker:
Slightly related: thinking about how we could write a test for this feature. I think currently, our test suite only supports checking for errors. Maybe I'll add @warn and @nowarn commands to the test application, where @nowarn would basically do -werror or something like that and @warn would check if the test assembled with a warning (rather than an error). Just an idea.

RPGHacker:
(Also whoops, did not mean to tag that username, haha, sorry)

KevinCathcart:
Yeah. I'll look into getting spc, as well as renaming the command, and getting the hex immediates to not trigger the warning, as mentioned in my previous email.

The tricky part of testing is that the cli test suite treats any output (even printing) as an error. So while making warn and nowarn work for the dll test suite is easy, doing it for the exe suite may be harder.

randomdude999:
I'm pretty sure that I "fixed" the CLI test suite so that it only marks a test as erroring if its stdout contains "error".

RPGHacker:
Yep, you did. I think otherwise xkasemu.asm would currently fail. That's why I would add some new command to the test suite, since expecting a test to throw a warning (or not throw a warning) seems like a reasonable case to test for me.

KevinCathcart:
Ah. I had not noticed that change to the app test suite. in that case enhancing it further sounds like the right approach.

randomdude999:
Yeah, the warn and nowarn special commands seem like a good idea. One thing that I don't really like about the test format right now though is that the test's special syntax conflicts with Asar's syntax, so it is necessary for the test program to remove the test lines and use a temporary file for assembling. IMO we should use `;!` as the start of the test lines, since that means it's not necessary to remove the lines from the patch to be applied. I think that's also the only thing that is done with the patches, so removing that would also eliminate the need for temporary patch files.

RPGHacker:
Good idea. Not sure why I didn't think about that earlier. In fact, those tests previously used to throw errors when ran directly in Asar, and it only worked after I made ;@ only throw warnings instead of errors on unknown command. Using an entirely different format seems way smarter here.

KevinCathcart:
Ok. Looking into it, it looks like the spc700 does not support any ambigious cases of 8 or 16 bit immediate values.

Now technically there are some cases where there exists ambiguity as to if an address (as opposed to an immediate) is direct page or not. Technically the same can occur with the direct page for the 65c816 architecture too. Technically : `LDA $F0+45` will wrap arround within the direct page, which is probably not what the user is expecting, especially if that $F0 came from a define. In other cases we could get a 16 bit address instead of the expected direct page address. That is often not an issue, since the longer instruction will still access the same memory.

However if the feature to remap the direct page is used (both the 65c816 and spc700 architectures can do that, although it is more limited in spc700), then this can become a serious issue.

Both are significant enough issues that a similar warning may be appropriate, but they also require more work to address, so are probably out of scope for now.

KevinCathcart:
I've now implemented not warning about hex constants. After all, nobody has any question that `LDA #$12` should be one byte, while `LDA #$1234` is two bytes.

That said, I've noticed that asar accepts hex constants with an odd number of digits. There is code in `getlen` to warn about these, but it is commented out:
```c++
			//if (i&1) warn(S dec(i)+"-digit hex value");//blocked in getnum instead
```
Unfortunately getnum does not actually block hex constants with an odd number of digits. 
I think we either want to block them, or reinstate that warning.

RPGHacker:
Didn't know that SPC700 and 65c816 were so different. I thought that since someone went through the trouble of adding .b/.w/.l support to it, there are probably cases of ambiguity.

Good question about three-digit hex numbers. I guess another option would be to always convert to the next-biggest, since a three-digit hex number will never fit into an 8-bit instruction, so. Not sure if that's the best idea here, though.

KevinCathcart:
Yeah there are definately cases of ambiguity for spc700, but just with respect to addresses as opposed to immediate values. I'd love to have some warnings for that, but I have personally been bitten by the immediates (especially those with calculations), and have not for addresses, and adding a warning for this required very little code change, as you can clearly see.

As for hex: I think currently the code does round up (i'll want to double check). I'm worried that code that looks like `LDA #$199` is likely a mistake. It could be that the user meant `LDA #$19`, but accidentally doubled a key, or wanted some other address and lost an input. Lastly it is certainly possible the user did want `LDA #$0199`, but dropping the leading zero is at best unconventional.

Personally I would favor restoring the warning globally for all hex numbers with an odd number of digits. 

However, if you don't want to go down that route, I would want to at least warn about it for simple hex immediates with the new `warn immediate on`, because the whole point is to warn about immediates that may be getting interpreted differently than the user intended.

RPGHacker:
I'm not entirely sure on this, either. On one hand, it sounds quite reasonable and useful. On the other hand, we would be enforcing coding style in the assembler, and I don't know if that is desired (for all we know, there might be people who like to use three-digit hex literals in certain situations). I guess it would be an easier decision if #13 was alrady implemented, in which case the warning could easily be disabled. Though enabling the warning just for `warn immediate on` seems like a good idea for the time being. After all, if the user specifically decides to enable some warning, they probably don't mind too much about an additional warning they get for free.

RPGHacker:
I've just merged the pull request. However, since the new warning system is now in place, I've made a small adjustment to it. Instead of `warn immediate on`, this warning is now enabled via the command `warnings enable W1013` (or by passing `-wW1013` on the command line). This is supposed to be the new standard way of enabling or disabling warnings, so I thought it would make the most sense (so that we don't have to introduce new commands for every optional warning we're going to introduce in the future). I've also documented the warning in the manual so that people will know how to enable it.

PR #84, opened 2018-04-28
Added opcode length support to spc700

### LINKED ISSUE #83: Opcode length support for SPC700 (specifically .b and .w)
I ran into an unusual situation while attempting to port my SPC700 code from TASM to ASAR...

I used the program counter to create a series of zero page variables (I also did one page variables, although I chose to mostly reserve them for the stack), like these (using the SPC registers as an example)...

```
org $0000F0

;SPC700 Registers defined...
SPC_TEST: ;No comment.
skip 1
SPC_CONTROL: ;Timer and ROM control
skip 1
SPC_DSPADDR: ;DSP Register ID
skip 1
SPC_DSPDATA: ;DSP Register Content
skip 1
SPC_CPUIO0: ;SPC<->SNES data byte 1
skip 1
SPC_CPUIO1: ;SPC<->SNES data byte 2
skip 1
SPC_CPUIO2: ;SPC<->SNES data byte 3
skip 1
SPC_CPUIO3: ;SPC<->SNES data byte 4
skip 1
SPC_AUXIO4: ;Free memory?
skip 1
SPC_AUXIO5: ;Free memory?
skip 1
SPC_T0DIV: ;8010hz timer divider
skip 1
SPC_T1DIV: ;8010hz timer divider
skip 1
SPC_T2DIV: ;64080hz timer divider
skip 1
SPC_T0OUT: ;8010hz timer output
skip 1
SPC_T1OUT: ;8010hz timer output
skip 1
SPC_T2OUT: ;64080hz timer output
skip 1
```

Then I attempted to write something to one of these registers via the accumulator, expecting two bytes instead of three. For example, this snippet of code (I have replaced one of my defines accordingly...

```
snd_clear_dsp:
	clrp					;Zero direct page flag.
	mov A, #$FF
	mov SPC_DSPADDR, #$5C			;Stop all channels first.
	mov SPC_DSPDATA, A			;KOFF will be cleared
	mov Y, #$7F				;later on, though.
	mov A, #$00
```

When I compiled the code, I expected...
$8F $5C $F2 $C4 $F3
Instead, I got...
$8F $5C $F2 $C5 $F3 $00

There is an ambiguity in the syntax for testing the SPC700 that means I have absolutely no clue how to force a label like this to go from $C5 $F3 $00 to $C4 $F3 (as in, force an 8-bit pointer without using DB as an impromptu replacement for a label). I would have a similar situation if I were to use setp and then briefly use the one page section...

This is what I resort to when I find these kind of bugs and I have no idea how to force the issue...

```
snd_clear_dsp:
	clrp
	mov A, #$FF
	mov SPC_DSPADDR, #$5C	
	db $C4					;mov dp, A opcode
	db SPC_DSPDATA
	mov Y, #$7F
	mov A, #$00
```

KungFuFurby:
Added the ability to use .b and .w with spc700.
This corrects an ambigous length case with the spc700 by being able to manually define the opcode length.

This pull request closes #83.

### DIFF
--- a/src/asar/arch-spc700.cpp
+++ b/src/asar/arch-spc700.cpp
@@ -176,6 +176,17 @@ bool asblock_spc700(char** word, int numwords)
 	else if (numwords==2)
 	{
 		int numwordsinner;
+		//Detect opcode length before continuing
+		int opLen=0; //In case of .b or .w, this overwrites auto-detection of opcode length
+		unsigned int periodLocCount=0;
+		do {
+			if (word[0][periodLocCount] == '.') {
+				opLen=getlenfromchar(word[0][periodLocCount+1]);
+				word[0][periodLocCount]='\0';
+			}
+			periodLocCount++;
+		} while ((opLen == 0) && (periodLocCount < strlen(word[0])));
+		if (opLen > 2) { error(0, "Opcode length is too long"); }
 		autoptr<char*> parcpy=strdup(par);
 		autoptr<char**> arg=qpsplit(parcpy, ",", &numwordsinner);
 		if (numwordsinner ==1)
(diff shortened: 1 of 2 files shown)


RPGHacker:
Thanks!
Could you also add a few test cases with the new length specifier to arch-spc700.asm? Always a good idea to have a couple of automatic tests for new features.

KungFuFurby:
I'm going to opt out of adding a test there... because neither the 65816 nor the SPC700 have an opcode length test in the tests folder (at least as far as I can see regarding the .b, .w, .l and .d prefixes, with .d being invalid on all cases so far since there are no 32-bit parameters supported). Therefore, it might make more sense for me to create a brand new one that tests those features on both architectures.

RPGHacker:
Sure, that works just as well. Both solutions are fine with me.

KungFuFurby:
You are receiving a bonus commit from me that contains a parser bugfix in arch-spc700. I caught this while implementing the feature, since opcode sizes are not supposed to be supported for MOV A,????+Y.

PR #87, opened 2018-04-29
Allow string argument passthrough

### LINKED ISSUE #56: Support pass-through of string parameters in user-defined functions
Basically, something like this should be possible:

```
function readfilenormalized(filename, pos)=readfile1(filename, pos)/255
```

Right now, it will throw an error because it thinks all parameters in user-defined functions are doubles. I think one simple way of fixing it could be to allow optional specification of data types in user-defined functions:

```
; pos defaults to ": double" here because nothing is specified.
function readfilenormalized(filename : string, pos)=readfile1(filename, pos)/255
```

### LINKED ISSUE #86: Bug evaluating function parameters
While working on a pull request for a different issue I noticed that the following code fails fails the assert statement, when obviously it should succeed:
``` asm
org $8000
function test(abc,ab) = abc+ab
assert test(1,2) == 3
```
Pull request that contains a fix will be sent shortly.

KevinCathcart:
Fixes #56 without changing the syntax.
The following code (based on the example in that bug) will simply work with this patch.
```
function readfilenormalized(filename, pos) = readfile1(filename, pos)/255
```
The approach is simply checking to see if a function argument is one of the parameters of the user function currently being executed. If it is pass it directly through to function being called. 

This also fixes #86, which I found while coding this. (The fix is the same as was previously used for a similar issue matching the names of the built in functions).

### DIFF
--- a/src/asar/math.cpp
+++ b/src/asar/math.cpp
@@ -453,6 +453,35 @@ extern chartabledata table;
 // RPG Hacker: Kind of a hack, but whatever, it's the simplest solution
 static char errorstringbuffer[256];
 
+#define maxint(a, b) ((unsigned int)a > (unsigned int)b ? (unsigned int)a : (unsigned int)b)
+
+// Check if the argument is simply a parent function parameter being passed through.
+// If so return the index of the parameter, otherwise return -1.
+// Update source if it is a parameter being pass through.
+int check_passthough_argument(const char** source)
+{
+	const char * current=str;
+	while (isalnum(*current) || *current == '_' || *current == '.') current++;
+	
+	size_t len=(size_t)(current-*source);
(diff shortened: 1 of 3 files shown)


randomdude999:
You should probably write some tests for this too.

KevinCathcart:
Yeah. I've not yet added any test cases for this. I'm guessing we would want to add the test case for the string pass-through to the v160features.asm test.

I'm not sure if there is a particularly good spot for the bugfix test, or if i should create a functest3.asm.

RPGHacker:
Yeah, I'd personally just add it to v160features.asm. Just one or two simple tests should do.

KevinCathcart:
I've added the passthrough test to v160features.asm, but added the bugfix regression test as a separate test file, because it did not feel right to put that in a features test file.

PR #92, opened 2018-05-02
Fix checksum being incorrect when rom doesn't yet have checksum

### LINKED ISSUE #91: Asar generates incorrect checksum when the output file has invalid checksum
(by output has invalid checksum, i mean that the checksum + the complement doesn't equal $FFFF)
This is because asar calculates the checksum by just adding the bytes of the rom together, but it doesn't make sure the checksum bytes add up before calculating. This could be fixed by overwriting the checksum with $0000 and the complement with $FFFF before calculating.

randomdude999:
Closes #91.

### DIFF
--- a/src/asar/libsmw.cpp
+++ b/src/asar/libsmw.cpp
@@ -505,6 +505,8 @@ bool goodchecksum()
 
 void fixchecksum()
 {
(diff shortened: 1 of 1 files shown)


randomdude999:
Whoops, I broke the sa1bankswitch.asm test - it still checks the ROM checksum. I'll fix that real quick.

PR #101, opened 2018-06-02
Memory file support

RPGHacker:


RPGHacker:
That's quite bizarre. So a reduction of coverage is treated as a fail?

randomdude999:
Apparently the default is any decrease makes it fail, but you can configure it [here](https://coveralls.io/github/RPGHacker/asar/settings#lnk_alerts).

RPGHacker:
Alrighty, will probably do so. Don't think it needs to fail over decreased coverage (since the coverage currently is really just there for informational purposes).

Anyways, before merging this, could you do me a favor? I want to have memory files covered by the test suite in at least some way. My idea here would be to work off the DLL test and add a slightly modified version of it which, before assembling anything, would recursively read all files in the test directory into memory to pass to the API as memory files. The entire contents of the directory aren't even a megabyte in size, so that should be no problem. We currently have no method of checking whether a file is actually read from memory by Asar (except for debugging, but that won't help us here, of course). Nevertheless, I would still like to have it covered by tests.

Thinking about it, I have another idea. We could also add a new optional warning to Asar which is thrown whenever Asar accesses a phyical file. We could then just enable it via the warning API in this new test and if the warning isn't thrown at all, we hopefully know that memory files are working properly.

randomdude999:
I had the idea of creating an entirely separate test suite that, instead of testing functionality in patches, separately tests the DLL functions, probably via a single patch file that has a few labels and defines and stuff. Then there would be no real need to run the main test suite twice (both with the DLL and the standalone executable).

randomdude999:
I have used these quite extensively in my project and there doesn't seem to be any problem with them (except for the fact that nothing tests them automatically but that's a different issue). I've also looked over the code and it looks fine. IMO this is good to merge.

RPGHacker:
Just posted this on SMW Central as well: my internet has spontanously disappeared a few days ago and I don't know when it will be back, so I won't be able to do much at home (currently using my internet at work). Since this requires a manual merge, I can't do it myself right now. Don't want to block progress on Asar, so for the time being, I'm putting you in charge. Feel free to merge this yourself and do other stuff on the repository while I'm unable to.

Yeah, I would prefer to have a test implemented before merging this, but it's not that urgent, so feel free to merge it even without a test.

RPGHacker:
```
virtual_filesystem::virtual_file_type virtual_filesystem::get_file_type_from_path(const char* path)
{
	if(m_memory_files.exists(path)) {
		return vft_memory_file;
	} else {
		return vft_physical_file;
	}
}
```

This won't do. Look at where that function is called. For example, in create_absolute_path() it's passed a relative path. One of the main reasons for having a normalize_path() function was to ensure that memory files work realibly and consistently. Checking them via a relative path defeats this purpose. I've also mentioned this problem in the original pull request by Horrowind.

Solution: the way create_absolute_path() works needs to be changed entirely. We can't really know if a file is a memory file or a physical file without creating the complete path first. create_final_path() needs to be adjusted for this. Instead of doing

```
if (path_is_absolute(test_path)) [...]
```

it needs to do something like

```
if (m_memory_files.exists(test_path) || path_is_absolute(test_path)) [...]
```

In create_absolute_path(), we could likely just drop the switch case entirely and end up with

```
string virtual_filesystem::create_absolute_path(const char* base, const char* target)
{
	if (is_path_absolute(target) || base == nullptr || base[0] == '\0')
	{
		return target;
	}

	return physical_file::create_final_path(target, base, m_include_paths);
}
```

(Might also be a good idea to change the `return target;` into a `return normalize_path(target);`- don't know if it can ever get a non-normalized path, but it's better to err ont he side of safety here)
is_path_absolute() also probably doesn't make much sense with the current implementation and should always do path_is_absolute(). Then we also need to add a create_final_path() to open_file() (the create_final_path() from physical_file::open() can then be removed).

I think that's all and hope I didn't miss anything.

RPGHacker:
I've added the suggested changes in 91450c03d0451c23107f9f05de7f6d4012067a4e. Of course we currently can't test whether they actually improve anything better. That's why I really wanted to have a test for this. I believe that a test would have failed in includehierarchy.asm and includeonce.asm, making us aware of the problem earlier.
(Basically, the test would run the DLL test, but throw an error whenever whenever a file was loaded from a location that was not in memory)

PR #108, opened 2018-06-22
[MinGW]Fix for cross-build 

boldowa:
# Summary

A compile error occurred when cross-building, so I fixed it.

## Changes

- Changed the include file name (**W**indows.h to **w**indows.h)
- Added settings and cmake files for travis.ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+sudo: false
 compiler:
   - clang
   - gcc
@@ -27,3 +28,29 @@ addons:
   apt:
     packages:
       - lcov
+
+# MinGW cross-builds (make only)
+matrix:
+  include:
+    - os: linux
+      compiler: i686-w64-mingw32-gcc
+      addons:
+        apt:
+          packages:
+            - binutils-mingw-w64-i686
+            - gcc-mingw-w64-i686
+            - g++-mingw-w64-i686
+      script:
+        - cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE=../cmake/mingw-x86.cmake $CMAKE_EXTRA_OPTS ../src
+        - make
+    - os: linux
+      compiler: x86_64-w64-mingw32-gcc
+      addons:
+        apt:
+          packages:
+            - binutils-mingw-w64-x86-64
+            - gcc-mingw-w64-x86-64
+            - g++-mingw-w64-x86-64
+      script:
+        - cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_TOOLCHAIN_FILE=../cmake/mingw-x64.cmake $CMAKE_EXTRA_OPTS ../src
+        - make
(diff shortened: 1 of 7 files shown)


randomdude999:
You override the APT packages that are installed, so the lcov package isn't installed in your added builds. Could you fix that (by just adding lcov to the list of apt packages to install?)

randomdude999:
Actually, it looks like you don't even run the test suite. In that case, update the line that sets GEN_COVERAGE to true, to exclude your builds too.

boldowa:
OK, I'm done.

Also I added cache settings. (.travis.yml : line 9-13)

.travis.yml
```diff
@@ -17,13 +25,45 @@ script:
   - make
   # run dll test suite
   # We need to pass the DLL path as an absolute path because it is used to calculate the std defines location, which will throw a warning when it's a relative path
-  - asar-tests/asar-dll-test $TRAVIS_BUILD_DIR/build/asar/libasar.* $TRAVIS_BUILD_DIR/tests ../dummy_rom.sfc $TRAVIS_BUILD_DIR/tests_dll_tmp
+  - asar-tests/asar-dll-test$EXE_SUFFIX $TRAVIS_BUILD_DIR/build/asar/libasar.* $TRAVIS_BUILD_DIR/tests ../dummy_rom.sfc $TRAVIS_BUILD_DIR/tests_dll_tmp
   # run app test suite
-  - asar-tests/asar-app-test asar/asar-standalone $TRAVIS_BUILD_DIR/tests ../dummy_rom.sfc $TRAVIS_BUILD_DIR/tests_app_tmp
+  - asar-tests/asar-app-test$EXE_SUFFIX asar/asar-standalone $TRAVIS_BUILD_DIR/tests ../dummy_rom.sfc $TRAVIS_BUILD_DIR/tests_app_tmp
```
randomdude999: You are missing a $EXE_SUFFIX here. This should say `asar/asar-standalone$EXE_SUFFIX` instead of just `asar/asar-standalone`

.travis.yml
```diff
@@ -17,13 +25,45 @@ script:
...
+      - lcov
+
+# MinGW cross-builds (make only)
+matrix:
+  include:
+    - os: linux
+      compiler: i686-w64-mingw32-gcc
+      sudo: true # Wine requires a lot of disk spaces. Therefore, use "Ubuntu Trusty VM" environment.
+      env:
+        - COMPILER=i686-w64-mingw32-gcc # $CC will overwritten by Travis.
```
randomdude999: Isn't $CC set to the value of the `compiler` option, which is also set to `i686-w64-mingw32-gcc`?

boldowa: $CC had gcc set. I don't even know why it'll happen.

RPGHacker:
I've lost track of this a bit. Is this ready for a merge already, or was something stil not working?

randomdude999:
I'm still not quite sure that the test suite actually runs the windows executables. I have an idea how to test it, but it would require merging this (into a separate branch at least).

PR #122, opened 2018-09-13
Optional outputrom

DevLaTron:
This patch adds the possibility to add an optional output name for the patched ROM:

**asar patch.asm original.smc patched.smc**

 will copy original.smc to patched.smc before processing. If no output name is provided, the previous behavior is conserved, making the patch backwards compatible by patching the ROM in place.

Technically, it simply copies to ROMNAME to OUTNAME and then sets ROMNAME to OUTNAME afterwards.

The rationale for this patch is that when using a Versioning System or automated build processes for patching, you may want to conserve the original file while outputting to a build directory or similar.

While there is the small drawback of having to press enter once more when using the prompt to patch, the advantages of easier automation should outweigh this by far.


trillllian:
feel old yet? this PR was created closer to the initial release of asar than to today

the ability to do this would be useful, but i don't like this approach, and the patch itself has bitrotted.

PR #123, opened 2018-09-13
Bugfix for segfault when including files

DevLaTron:
Using stat to check that we are including a regular file here.

This is a bit wonky, but well, a lot of the code here is.


randomdude999:
I implemented this myself (although in a slightly different way, check commit 606a983fc2a72e2716edd6d7ddd7c91c34688301), so closing this.

Issue #124, opened 2018-10-13
Add address-to-line mapping and source file list to WLA Symbols file


CypherSignal:
Recently, I submitted the following PR for wla-dx: https://github.com/vhelin/wla-dx/pull/202

Most notably, it adds extra information into WLA's Symbols file, such as a mapping of ROM addresses to source files and line numbers. In the PR, it is demonstrated how this information is used, including allowing VS Code to act as a moderately capable IDE for patch development, allowing for stuff like interactive breakpoints flagged in the editor, and reporting callstacks that refer to source files.

![VSCode displaying breakpoints, current PC, and callstack](https://i.imgur.com/4xFTpF9.png)

Given how much asar is used in the homebrew community for ROM and patch development, I wanted to look into developing similar support here, as well. I'm a bit unfamiliar with Asar and how it operates, but with a quick glance, it seems that support would be more involved than the changes needed for wla-dx and, hence, may be a greater burden to maintain in the long run (e.g. wla-dx already had a lot of support for mapping ROM addresses to source text by way of its List file output). Because of this, I want to, in particular, seek permission from the maintainers to develop this feature, or even, perhaps, advice on how it would be preferable to develop it.

So, would there be any thoughts or comments on inclusion of such a feature?

randomdude999:
A rom-address-to-source-line mapping is actually pretty simple - all the ROM writes should already go through a write1 function which can be hooked into, and i'm pretty sure the current line number is stored in a global somewhere (for error reporting). However, this approach would probably give every byte of the written area separately (i.e. `LDA #$64` will result in 2 different entries), which would make the list file absolutely massive. A smarter approach would probably be to use the existing written block tracking to figure out the blocks written in each line, maybe at the end of `assembleline`.

RPGHacker:
(Partially ninja'd by randomdude999, but here goes my reply, anyways)

Regarding permission: sure, feel free to go ahead. Don't think any of us could really prohibit this, anyways, since Asar's license permits modifying the source code for most purposes. It's also a very useful feature, so if you end up with a working solution and make a pull request, it's very likely we'll merge it into this repository (although it might take a while, because activity on Asar development is currently pretty low and likely won't change for at least a couple of months).

[...]

PR #125, opened 2018-11-24
Add address-to-line mapping and source file list to WLA Symbols file

CypherSignal:
[Implementation for task #124 ]

First off, sorry this took so long to set up in the face of something relatively simple :) 

Anyway, given the suggestions that were made, I found the best success in writing out addr2line mappings after the arch-specific block assembly. I tested this only against a couple large Asar projects, i.e. dotsarecool's smwdisx, and smwpracticecart, and it seems to work well, having been able to load up the output sym files into an emulator with the necessary support.

Note, compared to the changes done to wla-dx, the extended addr-to-line data is included by default in the wla symbols, instead of necessitating another cmd line argument. Given that wla-dx has so many more targets, and emulators reading in those symbol files, changing the files there would be a bit more dangerous. However, it seems that the emulators for SNES that read in wla symbols (in a non-compliant manner) seems to be almost nil, so it's not as much of an issue, perhaps?


randomdude999:
Just glancing over the code, it seems like you don't handle commands that write data (e.g. `db`, `fill`, `incbin`) at all. Would be great to have support for those too.

RPGHacker:
Can that even work? I thought the main purpose of the symbol file was to map instructions to code lines so that debuggers can use them, and incbin/db/fill etc. usually aren't used for instuctions.

CypherSignal:
That is correct, and somewhat intentional. Even putting aside the fact that the db and incbin data was one thing that blew up the size of the symbols in the cases I was looking at when I just tried out writing each byte, I'm not sure how useful it would be. The intent of the functionality is to map executable bytes in ROM to source lines, binary data being executable without being, say, opcodes themselves, seems a bit odd. At the very least, Labels still exist, for example, so a label can be dropped in ahead of binary data to locate the final ROM address of that stuff if necessary.

CypherSignal:
Hi, I was wondering if there was any follow-up desired for this, or if there are any blockers that I need to handle? The build-failure in appveyor seems to be independent of this change, and due to a configuration error, and aside from the issue of the final output file's checksum not being accurate, this seems to work (both testing on my machine on several different ROMs, and also from one other developer that was able to clone this work).

If the delay in this is just "We haven't gotten around to it", that's fine, of course; I just want to make sure that *is* the only delay, or something along those lines :)

RPGHacker:
Yeah, it's pretty much just that, neither randomdude, nor me, are currently working actively on Asar, and in my case I don't know when I will again (will probably take a while), so I haven't really had a deeper look at the pull request yet. Once I am actively working on Asar again and can take a good look at this, I might post a comment here if there's anything I want to talk about. As mentioned, though, it will likely take a bit of time still (at least for me, can't speak for randomdude, of course).

src/asar/libsmw.cpp
```diff
@@ -403,19 +404,33 @@ bool openrom(const char * filename, bool confirm)
...
+		{
+			unsigned char* filedata = (unsigned char*)malloc(sizeof(unsigned char) * (romlen + header * 512));
+			if (header)
+			{
+				fseek(thisfile, 0, SEEK_SET);
+				fread(filedata, sizeof(unsigned char), 512, thisfile);
+			}
+			memcpy(filedata + (header * 512), romdata, (size_t)romlen);
+			romCrc = crc32(filedata, (unsigned int)(romlen + header * 512));
+			delete filedata;
```
randomdude999: Things allocated with `malloc` should be freed with `free`. You should only use `delete` for things allocated with `new`.
```suggestion
			free(filedata);
```

src/asar/interface-lib.cpp
```diff
@@ -499,7 +499,7 @@ EXPORT mapper_t asar_getmapper()
 }
 
 EXPORT const char * asar_getsymbolsfile(const char* type){
-	symbolsfile = create_symbols_file(type);
+	symbolsfile = create_symbols_file(type, 0);
```
randomdude999: I don't really like how this just passes 0 for the checksum. I think this should just calculate the checksum under the assumption the ROM is unheadered.

src/asar/crc32.h
```diff
@@ -0,0 +1,10 @@
+/*
+* Generic crc32 helper function
+*/
+
+#ifndef __CRC32_H__
```
randomdude999: The rest of the header files use `#pragma once`, so I feel like this should too.

src/asar/assembleblock.cpp
```diff
@@ -963,7 +970,15 @@ void assembleblock(const char * block)
 		}
 	}
 	else if (numif!=numtrue) return;
-	else if (asblock_pick(word, numwords)) { numopcodes++; }
+	else if (asblock_pick(word, numwords))
```
randomdude999: Using this for checking opcodes is bad because for some reason, mapper changing commands (e.g. `lorom`, `hirom`) are also parsed in here. This means the addr-to-line mapping will contain the line numbers for mapper commands, which is probably not a very good idea.

src/asar/main.cpp
```diff
@@ -835,13 +838,53 @@ static void printsymbol_nocash(const string& key, unsigned int& address)
...
+
+		symbolfile += "\n[labels]\n";
 		labels.each(printsymbol_wla);
+
+		symbolfile += "\n[source files]\n";
+		const autoarray<AddressToLineMapping::FileInfo>& addrToLineFileList = addressToLineMapping.getFileList();
+		for (int i = 0; i < addrToLineFileList.count; ++i)
+		{
+			char addrToFileListStr[256];
+			snprintf(addrToFileListStr, 256, "%.4x %.8lx %s\n",
```
randomdude999: For me at least, the file CRCs get printed as 16 hex digits, with the 8 first ones being `ffffffff`. You could try casting the unsigned long to an unsigned int before printing.
(note: I think this only happens on 64-bit Unixes, where a long is 64 bits instead of 32.)

src/asar/assembleblock.cpp
CypherSignal: Hmm, perhaps. I can check into this. Handling superfluous data is not an issue, but if the same address points to two lines, that may not be ideal. Consumers of the symbol data should still handle such situations, but it may result in situations where a dev wonders why their PC is pointing at one line and not the other.

src/asar/main.cpp
CypherSignal: That, and the crc algorithm needs to use _strictly_ 32-bit data types as well. Someone on wla-dx noted that unsigned long is 64-bit on unix platforms (which was news to me 😬) so the size of those types need to be made stricter, too.

randomdude999: Yeah, I noticed that the hashes themselves didn't match up with the output of a crc32 tool. Maybe that's the cause then.

Alcaro: "a crc32 tool" - which one? There are many different crc32s, giving different results; [Wikipedia lists four different ones.](https://en.wikipedia.org/wiki/Cyclic_redundancy_check#Polynomial_representations_of_cyclic_redundancy_checks) I believe Unix crc32 uses CRC32-C, not _the_ CRC32 used in PNG/ZIP/etc.

also the only strictly 32bit data type is stdint.h's int32_t/uint32_t, quit fooling around with int/long. For printing, I'd recommend inttypes.h PRIx32, though it is annoyingly long.

randomdude999: I used the unix one, yes, but quick testing with python and an online crc tool (which shows the input's crc in a dozen different formats) says the unix utility does calculate the real one, not crc32c. But yes, probably best to just use uint32_t.

RPGHacker: AFAIK, an alternate way of handling print arguments is to just explicitly cast them when passed to snprintf(). As far as I'm aware, which types the format specifiers expect is handled by the standard (I think int for %x?), so just casting the args to the respective type while passing them to snprintf() should be enough (well, "enough to not make anything crash", at least - don't know if that will necessarily print anything meaningful).

Alcaro: > quick testing [...] says the unix utility does calculate the real one

I know I've seen something called crc32 that uses a silly one, and I'm pretty sure I've had bad results from that tool. Maybe I'm mixing it up with the SSE instruction, and accidentally gave /usr/bin/crc32 a SMC header, or something.

> As far as I'm aware, which types the format specifiers expect is handled by the standard

[Correct](http://www.cplusplus.com/reference/cstdio/printf/)

> (I think int for %x?)

Almost, it's unsigned int. And printf("%.8x", (unsigned)42) isn't much shorter than printf("%.8" PRIx32, 42).

> don't know if that will necessarily print anything meaningful

Converting an unsigned type to a smaller unsigned type will truncate it; converting to bigger will fill with zeroes. Too big type when printing is harmless (may waste a couple of clock cycles on some architectures, but I don't think long is slower than int anywhere that matters); too small won't happen, int is 32bit on everything anyone cares about except Arduino.

src/asar/assembleblock.cpp
CypherSignal: So, yeah, this will insert an entry, but I don't think it's an issue. Given that this will typically be the first item encountered, it'll write out an address of FF:FFFF for that file/line, and the consumer of the data will just never look that up in the map. Even if that's inline with other code, the extra line will be benign for all intents and purposes.

src/asar/main.cpp
CypherSignal: Updated things in commit 6c03fc0. On windows, it generates the same results, naturally, but with the types clarified, it should also generate the same data for a unix build.

PR #130, opened 2018-12-21
Fixing appveyor

CypherSignal:
The default vcpkg setup was frustrating some of the generated vcxproj files internal to cmake, accidentally referring to some lib files that didn't exist, causing c/cpp compiler identification to fail.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,16 @@ install:
 
     set PATH=C:\MinGW\bin;%PATH%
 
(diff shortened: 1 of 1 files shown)


randomdude999:
dude thank you so much for figuring out what was wrong with this

CypherSignal:
As a small note, the "on_failure" should illuminate problems like this should they occur in the future: https://ci.appveyor.com/project/CypherSignal/asar/builds/21177881 🙂

PR #136, opened 2019-02-02
Update C# module wrapper

VitorVilela7:
The C# Asar DLL/module wrapper was outdated compared to the other sources. It is now updated to the current API/Asar version.

I also did a minor code cleanup but it won't affect other tools that used the C# wrapper previously.


randomdude999:
Merging this until i figure out what i actually want to do with the dll api

PR #148, opened 2019-06-28
Change: Restore original sizeof syntax

p4plus2:
Change: Updated struct tests to reflect sizeof syntax
Bugfix: optimizeforbank not reset to default on pass start
Feature: Added dp and address optimizer for labels (and structs)
Feature: Added test for the new address optimizations
Feature: Added new c++11 initializer list constructor to assocarr
Cleanup: Rewrite the math function handler core

### DIFF
--- a/src/asar/asar.h
+++ b/src/asar/asar.h
@@ -96,6 +96,29 @@ extern int repeatnext;
 
 extern int optimizeforbank;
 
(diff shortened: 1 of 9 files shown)


p4plus2:
Forgot to remove some debug code.  will submit new PR when fixed.

PR #150, opened 2019-07-22
Rephrase a few things in the README.md

vince94:


RPGHacker:
Personally, I prefer the spelling "an SNES assembler", because I read it as "es en ee es", but I'm sure that's a debate that could be had for hours. :P

I personally agree with all of the other changes. Regarding the test format, since the section was already rewritten quite a bit, maybe it would be a good idea to rewrite it in a way that emphasizes the fact that tests are only really meant for the Asar repository? This first sentence here
```At the beginning of your asm files, you can write tests to ensure the correct values were written to the ROM after patching is complete.```
currently kinda makes it sounds as though it was a feature generally useful for ASM patches and not just for this repository.

**Comment by vince94** [comment]
I'll quick switch it back to "an"
Does "Please note that only Asar can run these tests, and other assemblers will treat them as normal comments" sound okay?

**Comment by Alcaro** [comment]
No. There are no Asar-compatible assemblers, and the tests are ignored by Asar too. They're only used by Asar's test suite (a program to test whether Asar still works, or if we accidentally broke something - we don't want to accidentally release something that emits wrong binary). Only the Asar developers should write those; people who just want to apply and create patches don't need to know they exist.

**Comment by vince94** [comment]
Ohhh I see! My apologies.

File: README.md [hunk 1]
```diff
@@ -24,7 +24,9 @@ Asar can also be built as a DLL. This makes it easier and faster to use in other
 * `tests` contains tests to verify Asar works correctly
 
 ## Test format
-At the beginning of your asm files, you can write tests to ensure the correct values were written to the ROM after patching is complete. (Intended for SMW, but there's also a dummy ROM included that should work with all tests.)
+*Please note that these tests are intended for use with Asar's test suite. Only contributers will need to use this functionality - people who just want to create and apply patches don't need to worry about it.*
```
Alcaro: > contributers

s/e/o

vince94: Oh jeez. Would you like me to fix that? I need to see if there's a spell check extension for notepad++ :/

Alcaro: [Thanks, but we already fixed that.](https://github.com/RPGHacker/asar/commit/c76c81a3e805b7c9e819aa3fb0123f22f99df560)

PR #171, opened 2020-09-11
[draft] fix / rewrite relative addressing

### LINKED ISSUE #170: relative addressing errors
First, apologies, this is an incomplete bug report. Working on a pull request, but, figured I'd put what I have here in case any 65816 gurus knew how to fix this quickly. I have some good repro cases though.

Bugs:
Instructions that use "relative addressing", like BRL [16bit] and BRA [8bit] can sometimes throw errors incorrectly for valid assembly. Here I'll use BRL as the example, but I think all this happens with BRA too.

This all seems to revolve around issues converting back and forth between 4 byte unsigned numbers and 2 byte signed numbers.

**Bug 1:** 

bug1.asm
```
ORG $C000DB
BRL label1          ; errors, but should generate 'BRL $B193'

ORG $C0B271
label1: BRK         ; doesn't matter, can be anything 
```

```
asar.exe bug1.asm
```

**Expected result**: Output ROM is:
82 93 B1
which is
BRL $B193

**Actual result**: Error message:
```
Errors were detected while assembling the patch. Assembling aborted. Your ROM has not been modified.
bug1.asm:3: error: (E5037): Relative branch out of bounds. (Distance is 45459). [BRL label1]
```

Analysis: Funny story: Asar actually does the right thing and outputs the correct bytes, but, afterwards a different bounds check fails. that check is located in https://github.com/RPGHacker/asar/blob/master/src/asar/arch-65816.cpp#L55 , in macro ```as_rel2()```. 

Here's that code with annotation:

int pos         = (int)num      -  ((snespos&0xFFFFFF)+3);
$0000B193  =   $00C0B271    -  ($00C000DB         + 3)

this section writes the bytes to the ROM:
```
write1((unsigned int)byte);  // $82 = BRL, this is correct 
write2((unsigned int)pos);   // $93B1, this is correct 
```

"BRL $B193" is interpreted the same as: "BRL -20,077"

so far so good. but now here's the error:
going into this, pass=2, foundlabel=true
```
if (pass==2 && foundlabel && (pos<-32768 || pos>32767))
    [error msg that prints 'pos']
```

specifically this check is the problem:
```
(pos<-32768 || pos>32767)
```

At first glance, the intention seems right.  The range of a signed 2-byte word which is 0x0000 thru 0xFFFF or -32768 thru 32767), so checking for it makes sense.

However, it causes this to incorrectly print:
```
bug1.asm:3: error: (E5037): Relative branch out of bounds. (Distance is 45459). [BRL label1]
```

If our ```pos``` variable really was a signed word (2 bytes) this would be fine because:
$B193 => interpreted as signed word =>-20,077, and that's in range of the bounds check.

-20,077 is the correct value we're looking for.

The issue is: ```pos``` here is actually a 4 byte unsigned int. In hex, its value is:
$0000B193

That _contains_ the right bytes (B193), but it's not quite good enough.  Interpreting this 4 byte unsigned int as a 4 byte signed int gives us an incorrect result of:
45,459

We're trying to get it to output -20,077. 

The hex representation of that is:
$FFFFB193 as signed int =  -20,077

Takeaways:
1. Using 4 bytes to hold the word that contains the offset is fine for writing the 2 bytes manually with write2()
2. The check can't look at the raw 4 bytes directly without either masking it with 0xFFFF, or, by converting the original value into it's signed form of $FFFFB193.  

That leads to the question of: what's the point of the check? If it's to validate the parsed input to make sure the user hasn't typed in a value > $FFFF, then, see below for the next bug:

**Bug 2.** 
If you pass in too large a value for BRL or BRA, asar will fail to throw an error but still write out the lower two bytes of the number.

bug2.asm
```
hirom
BRL $01B193
```

EXPECTED: fatal error thrown saying $01B193 is too large, max is $FFFF (2 bytes)
ACTUAL: no error thrown, and it outputs (wait.. twice in the ROM? seems even weirder) the following bytes:
82            93 B1
BRL          $B193

There's a check (same as the above one) that looks like it might catch this value being out of range, but, it's not called because we're not using a label here.

going into this, pass=2, foundlabel=false
```
if (pass==2 && foundlabel && (pos<-32768 || pos>32767))
    [error msg that prints 'pos']
```

we skip the:  ```(pos<-32768 || pos>32767)``` check, and continue with assembly, incorrectly not throwing an error.

So, we need to add a check that if ```pos > 0xFFFF``` then it throws an error.

-----------

More stuff to check, any bugfixes should add these as a test suite and need to make sure these work. (Disclaimer: from my reading of the CPU manual, I'm not super-familiar with 65816, so double check me)

expected (I think?) should throw an error that you can't BRL into a label in a different bank. even though the destination ROM address is less than 32k bytes away.
```
ORG $C9FF00
BRL label_in_another_bank
           
ORG $CA0000 ; should fail because branch can't cross bank. i.e. it can only modify PC, not DB
label_in_another_bank: BRK
```

That's only 256 bytes away in ROM, so it's in range of a 2 byte offset, BUT, [I think?] BRL is only for intra-bank transfers.

could add a check like:
```
	if (pass == 2 && foundlabel && ((snespos & 0xFF000) != (num & 0xFF000)))
                asar_throw_error(pass, error_type_block, error_id_bank_border_crossed, string("Can't branch to label in a different bank#."));
```

----

Other stuff:  
1. This is really my first time looking at asar, anything we need to keep in mind about ```freespace```? Does it use 0xFF000000 addresses or something? Do we need to deal with that in these checks?
2.  Need to check all this for scenarios where we're branching backwards to an earlier position in the bank (i.e. branching from address $C000F0 back to $C00000)

---

Anyway, I thought this might be easy but it turned into a rabbit hole :)

I was planning on writing some tests to cover all the cases.  It seems like if this gets more complex, we'll want to extract macros ```as_rel2()``` and ```as_rel1()``` into functions.  The only difference is as_rel1() operates on 1 byte like this:

https://wiki.nesdev.com/w/images/7/76/Programmanual.pdf
```
The Operand byte, a two’s complement signed value, is sign-extended to 16 bits, then
added to the Program Counter (its value is the address of the opcode following this one)
```

So, probably they could both call a new function like relative_address(offset, num_bytes), and it only needs to change the range check and # of bytes written.

Phew. OK, lemme know what you think.  Open to suggestions.

binary1230:
DRAFT, for review.  

This might actually be good to ship, I just want to throw some more tests and it, and maybe it run it on a few larger SNES assembly projects to make sure I didn't break anything. Disclaimer: I'm not an expert in SNES asm, just fixing stuff I have a good repro for.

Fixes #170 (details within)

Changes:
- Fix false positive out of bounds error message incorrectly firing for valid code
- Validate hex literals from assembly code to be in range (i.e. "BRA $FFFF" is more than the 1 byte limit, so flag it)
- Raise error if trying to jump to label outside our current bank (BRA/BRL can only change 16bit PC so can't change bank, they overflow PC)

Pre-merge checklist:
- [ ] Write some asm tests that mirror my unit tests
- [ ] Step through with a SNES debugger to verify accuracy on some tests
- [x] Check this on a large dissasmbly project for 1:1 binary correctness
- [ ] Should we do any of the checks outside of pass 2?
- [ ] I wrote this a little verbose, do you guys want me to condense more?
- [ ] Should I put this in a CPU agnostic file in the code (i.e. this might be also applicable for SPC). I haven't look at any of the other CPUs and if they implement things the same way. Guessing, probably.
- [ ] Test error message printout

Future me / future coders notes: 
- Make sure you really understand the guts of how signed 2 complements work :)
- The value of 'num' is 4 raw bytes, when casting it to anything less than 4 bytes you have to be super-explicit
- BRA can only jump -127/128 bytes. HOWEVER, we need to convert this value to 16bit signed value once we've read that byte of user input.  See the note in the BRA instruction in the SNES manual https://wiki.nesdev.com/w/images/7/76/Programmanual.pdf#page=308
- I wrote some unit tests outside of the project that test this function specifically, would be cool to include them somewhere but setting all that up is probably an entire other project.

### DIFF
--- a/src/asar/arch-65816.cpp
+++ b/src/asar/arch-65816.cpp
@@ -16,6 +16,45 @@ void asend_65816()
 {
 }
 
+// TODO: maybe rehome this to a less CPU-specific place, if it's the same for other arch's like spc
+//
+// is_long: if true, 16bit mode (i.e. BRL). if false, 8bit mode (i.e. BRA)
+void relative_addr(const unsigned int instruction, const unsigned int num, const bool is_long)
+{
+	int delta = (int)num;
+	if (foundlabel)
+		delta -= snespos + (is_long ? 3 : 2);
+
+	if (pass == 2) {
+		// TODO: need to check if we need to do any of this stuff on passes other than 2 -Dom
+		
+		if (!foundlabel) {
+			if (delta & ~(is_long ? 0xFFFF : 0xFF))
+				asar_throw_error(pass, error_type_block, error_id_invalid_input, 
+					"Relative address operand too large: " + hex4((unsigned)delta));
+
+			// Tricky:
+			// 1. Interpret our hex literal as a signed value, either 1 or 2 bytes based on is_long
+			// 2. Then, always store that result as signed 2 byte value
+			delta = (signed short)(is_long ? delta : (signed char)delta);
+		} else {
+			if (unsigned(snespos & ~0xFFFF) != (num & ~0xFFFF))
+				asar_throw_error(pass, error_type_block, error_id_bank_border_crossed, 
+					"Relative address: Label {TODO} must be in the same bank.");
+
+			if (!is_long && ((signed short)delta < -128 || (signed short)delta > 127))
+				asar_throw_error(pass, error_type_block, error_id_relative_branch_out_of_bounds, 
+					dec(delta).str);
+		}
+	}
+
+	write1(instruction);
+	if (is_long)
+		write2((unsigned)delta);
+	else
+		write1((unsigned)(signed char)delta);
+}
+
 extern bool fastrom;
 
 bool asblock_65816(char** word, int numwords)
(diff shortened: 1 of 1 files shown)


p4plus2:
>  Should I put this in a CPU agnostic file in the code (i.e. this might be also applicable for SPC). I haven't look at any of the other CPUs and if they implement things the same way. Guessing, probably.
Well, the SPC only supports short branches, but otherwise, this code could probably be reused there.  That said, I don't think its a worth while adventure unless there is a plan to simply gut and replace the entire mnemonic processing core.  I don't think anybody would disagree that the code in that section isn't exactly pretty or extendable...

As for the code, and this is sort of a problem in asar already, just keep in mind that this doesn't work well if you place labels in banks which are mirrors of each other.  Not necessarily a deal breaking issue, just something to keep in mind.  I don't see a need to condense too much, clean up the todo and it should be alright.  

From a pure visual glance I'd give this a sign off(with a small side note below), but I have NOT actually tested this and that needs to be done a bit more in depth before I'd suggest a full green light.  But, this certainly does seem to be an improvement.

Side note: pretty sure the lhs needs to be a `string` in the first error message to fix the mingw build.

binary1230:
haven't forgotten about this! just been working on a release of the tool that depends on this change, should be trying to finish this out soon.

binary1230:
hi folks, was finally revisiting getting this merged in. could I get whitelisted for appveyor, I tried running it on my own fork but there's some FTP upload stuff that's failing. I think I fixed the build error it was complaining about

PR #176, opened 2020-09-17
bindings: Fix strict aliasing issues on Windows

Alcaro:
```
$ wine gcc -O3 -Wall asardll.c
asardll.c: In function 'asar_init_shared':
asardll.c:66:34: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 # define loadraw(name, target) *((int(**)(void))&target)=(int(*)(void))GetProcAddress((HINSTANCE)asardll, name); require(target)
                                 ~^~~~~~~~~~~~~~~~~~~~~~~
asardll.c:118:21: note: in expansion of macro 'loadraw'
 #define loadi(name) loadraw("asar_"#name, asar_i_##name)
                     ^~~~~~~
asardll.c:123:2: note: in expansion of macro 'loadi'
  loadi(init);
  ^~~~~
[repeat about 20 times]
```

also I think defining your own NULL is undefined behavior, better leave it to libc.

### DIFF
--- a/src/asar-dll-bindings/c/asardll.c
+++ b/src/asar-dll-bindings/c/asardll.c
@@ -1,5 +1,5 @@
-
-#define NULL 0
+#include <stdbool.h>
+#include <stddef.h>
 
 #if defined(_WIN32)
 #	if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


RPGHacker:
```*((int(**)(void))&target)=(int(*)(void))```
Damn! What a hellish kind of cast wast that, anyways? I get dizzy just looking at it, lol.
Yeah, I'd say the new version is quite the improvement.

Alcaro:
It's nothing scary, it's just a pointer to a function pointer. Guess I wanted a oneliner back then; can't take pointers to temporaries, so can't memcpy.

No clue why it wasn't `*(FARPROC*)&target=(FARPROC)...`, though.

PR #179, opened 2020-10-01
Experimental

p4plus2:
Proposed optimizations and variadic macros.


src/asar/autoarray.h
```diff
@@ -151,19 +165,16 @@ template<typename T> class autoarray {
...
 		}
 	}
 
 	autoarray()
 	{
-		ptr = (T*)malloc(sizeof(T) * 1);
-		memset(ptr, 0, sizeof(T));
-		bufferlen = 1;
+		ptr = (T*)malloc(sizeof(T) * default_size);
+		memset(ptr, default_size, sizeof(T));
```
randomdude999, line 176: are you sure you changed the right argument here?

src/asar/interface-cli.cpp
```diff
@@ -77,6 +77,15 @@ void onsigxcpu(int ignored)
 
 int main(int argc, char * argv[])
 {
+	
+	//string test = "asdassdasdasdasd";
```
randomdude999, line 81: what are these comments about

src/asar/libsmw.cpp
```diff
@@ -104,23 +104,30 @@ static void addromwrite(int pcoffset, int numbytes)
...
 	addromwriteforbank(snesaddr, bytesleft);
 }
+#endif
 
 void writeromdata(int pcoffset, const void * indata, int numbytes)
 {
 	memcpy(const_cast<unsigned char*>(romdata) + pcoffset, indata, (size_t)numbytes);
-	addromwrite(pcoffset, numbytes);
+	#ifdef ASAR_SHARED
+		addromwrite(pcoffset, numbytes);
```
randomdude999, line 113: why not put the ifdef inside addromwrite instead of copying it to 4 places? compilers should be good enough at inlining functions that do nothing

randomdude999 (approved):
i can't really review the string optimizations properly on mobile but rest looks good, the comments are minor enough that i can fix them after merging

src/asar/interface-cli.cpp
p4plus2: I saw those post PR, it was for some initial tests I had forgotten about.  Post merge they can be removed easy enough.

src/asar/libsmw.cpp
p4plus2: Fair.  I think my logic was to show at the call site that its not used in the CLI version.  But one could just as easily argue not to repeat code in multiple places.  That can be changed.

src/asar/autoarray.h
p4plus2: probably not, actually from memory I don't even think the memset is needed at all.  But I'll probably fix that anyways until I can prove otherwise.  autoarray needs some work to fix its support for Non-pod types anyways...  All the changes there mostly stem from SSO not being copy safe with realloc or memcpy.

PR #192, opened 2020-12-09
don't sign-extend character array indices

katrinasm:
When `char` is signed, casting characters 0x80...0xff to `size_t` will produce huge indices (e.g. 0x80 -> 0xffffff80) which will usually lead to undefined behavior here. Observed result is that Asar generates nonsense when fed "negative" characters in a string literal. Casting through `unsigned char` avoids this.

I have avoided touching the `const_cast` here because it doesn't seem to be necessary but seems to have been added to please a compiler.

### DIFF
--- a/src/asar/assembleblock.cpp
+++ b/src/asar/assembleblock.cpp
@@ -1252,10 +1252,10 @@ void assembleblock(const char * block)
 					asar_throw_warning(0, warning_id_xkas_patch);
 				for (char * str=const_cast<char*>(safedequote(pars[i]));*str;str++)
 				{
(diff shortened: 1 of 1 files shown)


randomdude999:
terribly sorry for not merging this earlier, i just didn't notice it, lol. it'll be in the next release which is planned soon-ish (i wanted to wait until p4 implemented global labels, i'm not sure what is going on with that though), though we might as well do it now and add global labels in the release after that. in the meantime, you can use the beta builds from https://random.muncher.se/ftp/asar for testing (the recommended windows binaries are [here](https://random.muncher.se/ftp/asar/windows/win32/build/asar/MinSizeRel/))

PR #196, opened 2021-01-16
Implement global labels (actual impl by p4plus2)

Atari2:
This PR implements global labels, the C++ code is all written by @p4plus2 so I take no credit for that, while the test cases were written by @Atari2 (me).


RPGHacker:
Let's not forget to also add the new fature to the manual.

PR #199, opened 2021-02-17
Improvement: Add support for 16bit constant from two ascii chars

furious:
Minor changes to add support for 16bit values made of two chars, eg.: `lda.w #'AS' : sta $00 : lda.w #'AR' : sta $02`

### DIFF
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -1529,6 +1529,9 @@ lda #~$80	; Equal to lda #$FFFFFF7F</code></pre>
 		<pre><code class="65c816_asar">lda #'a'
 sta $00
 
+lda.w #'bc'
+sta $01
+
 db 'x','x'+1,'x'+2</code></pre>
 		<h4 id="opcode-length">Opcode Length Specification</h4>
 		By appending <code>.b</code>, <code>.w</code> or <code>.l</code> to an opcode, you can specify that opcode's length. This is recommended in cases where the length could be ambiguous.
--- a/src/asar/asar_math.cpp
+++ b/src/asar/asar_math.cpp
@@ -734,8 +734,12 @@ static double getnumcore()
 	}
 	if (*str=='\'')
 	{
-		if (!str[1] || str[2] != '\'') asar_throw_error(1, error_type_block, error_id_invalid_character);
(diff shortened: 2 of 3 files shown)


Alcaro:
```
'a' = $1234
'b' = $5678
dw 'a','b'
```
yields, with the current Asar, 34 12 78 56.

What do you propose would be the correct output for the following snippets, and is that the actual output of this PR?
```
'a' = $1234
'b' = $5678
dw 'ab'
```
```
'a' = $12
'b' = $34
'c' = $56
dl 'abc'
```

furious:
I was disassembling a firmware that checks strings in specific locations, eg.:
```asm
; checking for "SUPERUFO" at 808000
org $808000:
    db "SUPERUFO"
...
    lda.l $808000
    cmp.w #$5355 ; loading as 16 bit 'US' instead of the actually readable 'SU'
    bne .not_superufo
    lda.l $808002
    cmp.w #$5045 ; 'EP' > 'PE'
    bne .not_superufo
    lda.l $808004
    cmp.w #$5255 ; 'UR' > 'RU'
    bne .not_superufo
    lda.l $808006
    cmp.w #$464F ; 'OF' > 'FO'
    bne .not_superufo
```

My intention of this PR is to make it more easy to work with 16 bit chars/text...

I was manually writing the opcode with the inverse char order, eg.: 
```asm
db $C9, "US" ; = cmp.w #$5355
```

furious:
Sorry, I was wrong, that piece of code I showed was written with the bytes actually flipped, so I just pushed a correction...

Here's a real example of this feature:
image
In this code, it checks for a specific header string above
image

p4plus2:
I don't really know how I feel about this one.  Maybe this is better as a builtin function (say `be` for big endian) so it can be used in any math context.  `be($1234)` or `be("US")` could then be supported for example.

I'm open to other suggestions too, but I'd rather not add a context sensitive hack.  A `be` function could hypothetically support strings larger too, up to four bytes.  Thus would work with `d[bwld]` (though db its effectively a noop)

trillllian:
too niche to be useful, also i think you can just implement this as a function in asar2 anyways

PR #200, opened 2021-03-06
Implemented function to convert RGB colors to RGB555 value

furious:
I've been doing it with custom functions but wanted to make another contribution to Asar.

This PR adds a new builtin function called "rgb555" (sugestion), that you can easily convert colors from hexadecimal `rgb555($FF00FF)` or like this `rgb555(255, 0, 255)`

That's it... =)

### DIFF
--- a/docs/manual/index.html
+++ b/docs/manual/index.html
@@ -2627,6 +2627,12 @@ max(a, b)</code></td>
 				<td><pre><code class="65c816_asar">lda #bank(some_label)</code></pre></td>
 			</tr>
 			<tr>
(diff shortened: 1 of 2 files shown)


p4plus2:
Not a bad idea but I believe this algorithm is wrong.  That will need corrected before we can accept.  Tests wouldn't hurt either.  But I can add those if you don't.

furious:
I've compared with some existing code to compare, feel free to add some tests =)

randomdude999:
i don't think this belongs in asar, you can easily implement this yourself as a custom function.

PR #201, opened 2021-03-08
Added strlen() builtin function + docs for it

Atari2:
Added a builtin function that given a string returns its length (simply doing string.length() internally).
I didn't add tests because it was fairly trivial to implement and I did some testing myself and everything appeared to work fine.
In the meantime I also updated the .gitignore file adding all the garbage that Visual Studio/CMake create upon building the sln file and the project.


Atari2:
Git screwed me over by being on multiple PCs and now I have to close and re-open the pull request. Sorry. Going to reopen it later.

RPGHacker:
Ah, alright, fair enough.

PR #202, opened 2021-03-09
Added strlen() builtin function + docs

Atari2:
Added a builtin function that given a string returns its length (simply doing string.length() internally).
I didn't add tests because it was fairly trivial to implement and I did some testing myself and everything appeared to work fine.
In the meantime I also updated the .gitignore file adding all the garbage that Visual Studio/CMake create upon building the sln file and the project.

### DIFF
--- a/src/asar/assembleblock.cpp
+++ b/src/asar/assembleblock.cpp
@@ -1631,7 +1631,7 @@ void assembleblock(const char * block, bool isspecialline)
 		{
 			if (fixedpos && freespaceorgpos[freespaceid]==-1) return;//to kill some errors
 			snespos=(freespaceid<<24)|freespacepos[freespaceid];
-			resizerats(snespos, freespacelen[freespaceid]);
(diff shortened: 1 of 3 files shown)


RPGHacker:
Whoops, kinda late in commenting on this.

It definitely makes sense to add a function like this, although I'd like to propose a name change. Since all of our other string functions have a more explicit name, I'd prefer something like "stringlength()", even though that's longer.

In fact, I think even stringlength() won't be enough, because it's very likely that UTF-8 support will be added to Asar at some point, and for that reason, if we add a string length function now, I'd like its purpose to be made clear and explicit right away. Therefore, I'd personally propose the name "stringbytelength()" (so that if we do add UTF-8 later on, we can then have a stringcharlength() function). I know that's quite long, but I'm also not sure how often string functions will be needed in Asar, anways, so I personally don't mind that.

Atari2:
Closing this for now, since it's likely to not get merged anytime soon.

PR #220, opened 2021-09-20
Add invalid branch distance error checking on BBC/BBS opcodes

### LINKED ISSUE #219: BBC/BBS opcode in SPC700 architecture is not reporting an out of range error when branch distance goes too far
Accidentally ran into this while working on my SPC700 code thanks to me looking up the raw hex data. There appears to be no overflow checking in the BBC/BBS opcodes at all branching distance-wise.

Here's the exact lines that has the fault:
https://github.com/RPGHacker/asar/blob/3553e3abca02c672d96f956ce87edb39545d7652/src/asar/arch-spc700.cpp#L359-L367

KungFuFurby:
This merge request closes #219.

### DIFF
--- a/src/asar/arch-spc700.cpp
+++ b/src/asar/arch-spc700.cpp
@@ -356,6 +356,10 @@ bool asblock_spc700(char** word, int numwords)
 					write2(((unsigned int)bits<<13)|num);
 					return true;
 				}
+				int pos = (getnum(arg[1])- (unsigned int)(snespos)-3);
+				if (pass && foundlabel && (pos<-128 || pos>127)) {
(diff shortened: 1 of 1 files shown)


KungFuFurby:
If you're wondering why my build failed on the pull request, clang was the cause, and a timeout error seems to have been the cause (specifically, while uploading `asar.dll`): `Error uploading artifact to the storage: Unable to read data from the transport connection: The connection was closed.`

trillllian:
note: i deleted the foundlabel check your commit had. if the subtraction is unconditional, the overflow checking should be too. whether it should be unconditional is a different issue...

PR #228, opened 2021-12-30
Add documentation about error/warnings IDs along with the error message

### LINKED ISSUE #156: Document and hardcode error/warning IDs
error and warning ids are supposed to be unchanging. and it's supposed to be easy to determine what an error is from its code. currently, the only thing that guarantees the ids being static is their order in the source code. this makes it both easy to accidentally change the ids of some warnings/errors and very annoying to figure out what a specific id refers to without having access to some code that throws it. we should document all currently used error and warning ids in the manual, and also change the code so the ids are right next to the descriptions.

Atari2:
This PR adds a small change to the docs to add 2 tables that have the association ID->name->small description for both errors and warnings.
I found myself needing this more than once so I thought might as well add it.

The tables are auto-generated from asar's source code. They can be improved later (and they should), but this is a base stub that works and has all the information someone might need.

This could close #156 but that's up to the maintainers.


RPGHacker:
This could be useful. I think I recall someone asking for this, but I don't quite remember who or where.

I have a few suggestions:

- If the lists are auto-generated, then whatever was used to generate the lists (I assume a script?) should also be commited. Otherwise, I'm almost certain no one is going to maintain these lists in the long run.
- For the errors list, I think we can hide the internal name. These don't really appear anywhere publically in Asar (unless there's been plans to change that, like with warnings? I don't remember).
- For warnings, on the other hand, I think only the "internal" name should be shown, since we're actually deprecating warning IDs. (Talking about that, @p4plus2 I think Asar 1.9 and 2.0 currently both still print warning IDs when a warning is thrown. That feels a bit problematic when our plan is to deprecate them. I think we should change this to print warning names instead, which would require fixing the test suite + tests. Also, I don't remember if warn enable/disable already work with warning names. Anyways, since warning IDs are already being deprecated, I do think including all this in v1.9 is crucial.)

Atari2:
When I say "auto-generated" I mean:
 - I copy pasted the enum init code from errors.cpp/warnings.cpp from asar into a text editor
 - Did some manual text editing
 - Generated the html with python directly from the command line (I didn't use a script)
 
 I kept the internal name for both for consistency.

Atari2:
@RPGHacker Added a script to automate the generation of the HTML tables.

RPGHacker:
Thanks, this definitely looks a lot more maintainable.
I'll check it out in more detail later and then see if I can merge it.

docs/manual/warnings-list.html
```diff
@@ -0,0 +1,77 @@
...
+<tr><td><code>W1024</code></td><td>check_memory_file</td><td>Accessing file '%s' which is not in memory while W%d is enabled.</td><td>False</td></tr>
+<tr><td><code>W1025</code></td><td>if_not_condition_deprecated</td><td>'if !condition' is deprecated - please use 'if not(condition)' instead.</td><td>True</td></tr>
+<tr><td><code>W1026</code></td><td>function_redefined</td><td>Function '%s' redefined.</td><td>True</td></tr>
+<tr><td><code>W1027</code></td><td>datasize_last_label</td><td>Datasize used on last detected label '%s'.</td><td>True</td></tr>
+<tr><td><code>W1028</code></td><td>datasize_exceeds_size</td><td>Datasize exceeds 0xFFFF for label '%s'.</td><td>True</td></tr>
+<tr><td><code>W1029</code></td><td>mapper_already_set</td><td>A mapper has already been selected.</td><td>True</td></tr>
+<tr><td><code>W1030</code></td><td>feature_deprecated</td><td>DEPRECATION NOTIFICATION: Feature \"%s\" is deprecated and will be REMOVED in the future. Please update your code to conform to newer styles. Suggested work around: %s.</td><td>True</td></tr>
```
randomdude999, line 72: nice blackslashes

Atari2: thanks

randomdude999:
this is less useful now that we use the internal names instead of numeric IDs, but i guess it might still be nice to have. but the script needs to be fixed now that we don't use the numbers anymore.

also i think there are a few unused warnings and errors (in asar 2 at least). should review those aswell

randomdude999:
i updated the script myself and reworked it a bit, will commit shortly

PR #231, opened 2022-01-25
Include 'autoclean ...' lines in AddrToLineInfo

spooonsss:
Previously, lines that start with "autoclean" do not an entry in e.g. the `.wla` file's `[addr-to-line mapping]` section. This fixes that issue.


randomdude999:
good catch, thanks. i wonder if there are other commands that should emit addr-to-line info but don't... looks like db/dw/etc don't emit it, but i remember arguing with someone over whether they should or not, so i'll leave it up to others to decide

PR #238, opened 2022-04-09
Update index.html

Alcaro:


randomdude999:
looks like rpg hacker "fixed" this in the other direction at some point (making it throw an error, which imo is a dumb example). fixed it again.

PR #241, opened 2022-04-23
Fix asar-exe-test deadlock because write pipe is full

Atari2:
Previously the created pipe had a size of 4096, this lead to a deadlock on the xkasemu.asm test because it output 4400~ bytes of output, which, since not read from the other side, caused the pipe to become full and stop accepting input, blocking everything. 

This patch fixes it by simply using pageSize * 4 as a size, which is 4096 * 4 bytes. If a test surpasses 4096 * 4 bytes of output, it'll lock up again, for now this is not the case, however in the future a "better" fix would be to read from the pipe continously instead of at the end, as to not let it get full.


Atari2:
cc. @RPGHacker

PR #242, opened 2022-05-04
Don't removerats twice when two autocleans point to the same freespace

### LINKED ISSUE #216: Wrong address when cleaning freespace blocks multiple times
I came across this in GPS where for some reason two label in different freespace blocks had the same address. The simples test case I could find is creating a rom with
<pre>
check bankcross off
org $008000 : rep $80000 : db $00
</pre>

and then patching
<pre>
print "previous test1: $",hex(read3($008000+1))
print "previous test2: $",hex(read3($008004+1))
print "new test1: $",hex(test1)
print "new test2: $",hex(test2)

org $008000
autoclean JML test1

org $008004
autoclean JML test2

freecode
test1: db 1

org $008008
autoclean JML test1

freecode
test2: db 2
</pre>

twice which prints the following:

<pre>
first run:
previous test1: $0
previous test2: $0
new test1: $908008
new test2: $908011

second run:
previous test1: $908008
previous test2: $908011
new test1: $908008
new test2: $908008
</pre>

on the second run, both test1 and test2 have the same address even though they're supposed to be in separate freespace blocks.

edit: I should probably specify that this only happens if you one autoclean is before the freespace command and one after and then another freespace is requested.

Playing around with this a bit more: If test1 is cleaned, then requested then another freespace block is requested and then test1 is cleaned again asar throws an error ` (E5136): Don't autoclean a label at the end of a freespace block, you'll remove some stuff you're not supposed to remove. [autoclean JML test1]`

spooonsss:
Fixes #216

No test added. Since this bug requires existing RATS tags to repro, I don't think the current testing framework would support a test for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ These two characters should precede each test line, so that Asar sees them as co
 * 2 hex digits - a byte for it to check for 
   * You can specify more than one, like in the examples below, and it will automatically increment the offset.
 * A line starting with `+` tells the testing app to patch the SMW ROM instead of creating a new ROM
+* `#` followed by a decimal number tells the testing app to apply the patch that number of times repeatedly
 * `errEXXXX` and `warnWXXXX` (where `XXXX` is an ID number) means that the test is expected to throw that specific error or warning while patching. The test will succeed only if the number and order of errors and warnings thrown exactly matches what's specified here. Be wary that Asar uses multiple passes and throws errors and warnings across multiple of them. This can make the actual order in which errors and warnings are thrown a bit unintuitive.
 
 In addition to the format mentioned above, it's also possible to check for user prints a patch is expected to output (by `print`, `error`, `warn` or `assert` commands). This is done by starting the line with one of the following sequences:
--- a/src/asar-tests/test.cpp
+++ b/src/asar-tests/test.cpp
@@ -245,7 +245,7 @@ static void delete_file(const char * filename)
 // RPG Hacker: Original test application used system(), but
 // that really made my anti-virus sad, so here is a new,
 // complicated platform-specific solution.
-// NOTE: No cont char*, for commandline, because CreateProcess()
+// NOTE: No const char*, for commandline, because CreateProcess()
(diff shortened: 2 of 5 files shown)


Alcaro:
From a cursory glance, this looks like it'll just trade one bug for another, and leak freespace in cases like 
```
;patch1.asm
org $008000
autoclean jsl hijack1
autoclean jsl hijack2

freecode
hijack1:
rtl

freecode
hijack2:
rtl

;patch2.asm
org $008000
autoclean jsl hijack1
autoclean jsl hijack2

freecode
hijack1:
rtl
hijack2:
rtl
```
or, of course, things like autoclean dl or autoclean read3($00f606) (this patch only touches the autoclean jml/jsl handler).

If I remember correctly, autoclean originally ran in the first pass, and the freespace finder was second; can't delete something that's not there yet. Did we leave that model? If yes, was it due to bugs involving lda ForwardLabel : autoclean jsl AnotherLabel?

Proposed proper patch: For each autoclean, loop the list of freespaces created in this patch prior to this point. If that list contains this address, the autoclean target was already deleted and reused, and should not be deleted again.

RPGHacker:
> No test added. Since this bug requires existing RATS tags to repro, I don't think the current testing framework would support a test for this.

Actually, I think there is a way to tell the test suite to a apply a patch twice to the same ROM. We just probably forgot to document it in README.me.

Taking a peak at prot.asm, the syntax seems to be:
```;`+ #2```

spooonsss:
Alcaro is right, that testcase shows a bug introduced. And I forgot to handle `autoclean dl`. I'll rework this.

Alcaro:
Looks good to me, other than the nitpick. Good job wrangling this thing, Asar is a messy tool and this isn't its easiest part.

randomdude999:
i deleted this entire fix in asar2.0, since there rats tags are cleaned during pass 1 and freespaces are allocated at the end of pass 1. which i'm pretty sure avoids the root cause of this bug

PR #243, opened 2022-05-09
Build asar as a static library too. Add a way to optionally disable building the test app.

Atari2:
With these changes, building asar now also builds a static library version of it.
The name of the library will be asar-static.lib/libasar-static.a .
This behavior is enabled by default (as everything else) and can be disabled.
To do this, add -DASAR_GEN_LIB=OFF to the cmake configuration command.
Additionally, this improves the build configuration on MSVC using some cmake generator expressions.
It is also now possible to disable the build of the test app with -DASAR_TESTING_DISABLED=TRUE.
Note that the build of the test app is enabled by default, to keep the previous behavior.


RPGHacker:
Wasn't disabling tests already possible via the ASAR_GEN_EXE_TEST and ASAR_GEN_DLL_TEST variables? CMakeLists.txt of asar-tests still checks for those variables.

Atari2:
Yeah but I thought providing a single option to just straight up not process the CMakeLists.txt for the test was faster and shorter. If that's too much I can remove that.

RPGHacker:
It's not really a problem, no. Though I'm not sure if it actually has the desired effect.

Doesn't CMake parse the entire thing once, anyways, so that it knows which options even exist? Or is that only when using CMake GUI?

Atari2:
I don't use CMake gui so I have no idea but afaik when invoked from command line CMake just loads the CMakeLists.txt that's present in the folder that was specified, so like `cmake src` will only load the CMakeLists.txt present in src/ that will *then* load the rest. So if you prevent the `add_subdirectory` with an `if`, I believe that will avoid the whole processing of the new subdirectory entirely.

Atari2:
Also it might be worth adding a top-level CMakeLists.txt instead of having one in src/ because that way other people who use CMake can include asar in their project by using [FetchContent ](https://cmake.org/cmake/help/latest/module/FetchContent.html) which doesn't work if the CMakeLists.txt isn't in the top level folder of the github repo. This is just an idea thought, not sure if worth pursuing but it would be super easy to add.

Atari2:
Any news on this @RPGHacker ?

RPGHacker:
In terms of merging this? Nope, not really.

Motiviation and time for hacking-related stuff comes in bursts for me, and unfortunately I'm currently pretty low on both, and also not really in the right headspace. I'm guessing it's a similar story for p4 and randomdude999.

Is it urgent? Do you require this for something specific?

Atari2:
> Motiviation and time for hacking-related stuff comes in bursts for me, and unfortunately I'm currently pretty low on both, and also not really in the right headspace.

I feel you.

> Is it urgent? Do you require this for something specific?

It's not urgent but I require this because I would really like to stop distributing pixi with the asar.dll and instead link it in as a static library, which would help in multiple ways. I could monkey patch it but I really don't want to keep an asar fork up just to have asar as a .lib so having this merged into master would make it really easy.
It's fine if you don't want to look at this, I completely understand, I think p4 mentioned to me that he was going to take a look at it sometime this week hopefully.

RPGHacker:
I see. Yeah, in that case, maybe try bugging p4 or randomdude999. If they don't have the time, I could also grant you collaborator access to the repository so you could add those changes yourself.

There is actually one potential issue that I see with statically linking Asar into PIXI, though. In typical SMW hacking fashion, PIXI currently doesn't seem to specify any license. I'm no legal expert, but as far as I understand, that technically makes it incompatible with open source licenses. Since Asar is licensed under the LGPL, any application statically linking to it would also have to at least use LGPL. Currently, that's not the case with PIXI.

The best thing here would be to retroactively add an LGPL license to PIXI. However, I'm not sure if anyone but JackTheSpades is legally allowed do that - unless maybe we find some old comment/post by them renouncing any rights to PIXI.

Atari2:
> I see. Yeah, in that case, maybe try bugging p4 or randomdude999. If they don't have the time, I could also grant you collaborator access to the repository so you could add those changes yourself.
> 
> There is actually one potential issue that I see with statically linking Asar into PIXI, though. In typical SMW hacking fashion, PIXI currently doesn't seem to specify any license. I'm no legal expert, but as far as I understand, that technically makes it incompatible with open source licenses. Since Asar is licensed under the LGPL, any application statically linking to it would also have to at least use LGPL. Currently, that's not the case with PIXI.
> 
> The best thing here would be to retroactively add an LGPL license to PIXI. However, I'm not sure if anyone but JackTheSpades is legally allowed do that - unless maybe we find some old comment/post by them renouncing any rights to PIXI.

Asar ships with 3 licenses, which one is the correct one? Because if I go by https://github.com/RPGHacker/asar/blob/master/license-wtfpl.txt then I'm allowed to do what I want.

RPGHacker:
The relevant file is [this one](https://github.com/RPGHacker/asar/blob/master/LICENSE), btw. It explains in detail which of the other licenses applies to which situation.

Somewhat confusing, I admit, but that's just licensing in a nutshell.

Atari2:
After this commit 96265eafeba886c7d9bd92afdc910a5f64bced4d now pixi has a license compatible with asar, so this PR should be looked at again, maybe it's worth it :)

randomdude999:
okay but can you please fix the diffs (they show up as having changed every single line in the files you touched, impossible to review)

Atari2:
I have no clue what happened to the diffs. I'll just checkout master and redo the changes, they're simple and trivial so it should be no problem

Alcaro:
Looks like someone changed every LF to CRLF, or other way round.

You can get a sane diff at https://github.com/RPGHacker/asar/pull/243/files?w=1, but better fix the patch anyways. It'll screw up git blame and stuff, and we presumably want the linebreaks to keep their current style.

PR #247, opened 2022-06-18
Set PDB_NAME to something other than asar

Atari2:
CMake was automatically setting the PDB_NAME to be asar.pdb, which on Windows conflicts with the dlls pdb name, since it's the same (asar.dll -> asar.pdb). This was giving arcane linker errors since it was trying to write the same file at the same time from 2 places and it completely broke debugging on Visual Studio, since it wasn't loading the debug symbols. Setting the PDB_NAME alone works fine, symbols load correctly and there are no linker errors. 
This was also tested on Linux to confirm that it works just fine, cmake just ignores the property but doesn't complain.

### DIFF
--- a/src/asar/CMakeLists.txt
+++ b/src/asar/CMakeLists.txt
@@ -209,7 +209,7 @@ if(ASAR_GEN_EXE)
 	endif()
 	
 	set_target_properties(asar-standalone PROPERTIES OUTPUT_NAME "asar")
-	set_target_properties(asar-standalone PROPERTIES PDB_NAME "asar-standalone")
+	set_target_properties(asar-standalone PROPERTIES PDB_NAME "asar.exe")
 endif(ASAR_GEN_EXE)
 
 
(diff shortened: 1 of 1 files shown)


RPGHacker:
Oh, whoops! Of course the rename had to break something! 🙃
But does that really work? I thought Windows always just looked for a .pdb file with a matching name as the .exe?

Atari2:
Yeah it works just fine 
image

RPGHacker:
Interesting, I never knew. 👀
I suppose the compiler bakes the .pdb file name into the .exe then or something.

Anyways, could we adjust the name slightly? I'm thinking of

```set_target_properties(asar-standalone PROPERTIES PDB_NAME "asar.exe")```

for the executable and

```set_target_properties(asar PROPERTIES PDB_NAME "asar.dll")```

for the lib. That should give us file names like asar.exe.pdb and asar.dll.pdb, which should hopefully avoid any confusion. I think I kinda prefer that over having one of the .pdb files just named asar.pdb.

If you do that, I'll merge the PR (or alternatively I can just quickly do it myself, if you prefer that).

Atari2:
For some reason using:
`set_target_properties(asar-standalone PROPERTIES PDB_NAME "asar.exe")`
still produces asar.pdb. Maybe CMake strips .exe, which would make sense, considering it uses the filename of the binary originally.

Atari2:
Done :)

RPGHacker:
Neat, thanks!

PR #249, opened 2022-06-20
Reduce warnings

Atari2:
This patch makes the number of warnings on the MSVC build of asar go from 2.8k to ~160. 

The main reducer is going from /Wall to /W3. Since /Wall is like clang's [-Weverything](https://clang.llvm.org/docs/UsersManual.html#diagnostics-enable-everything), it tends to be overly noisy and warn about unimportant things, /W3 is much less noisy while still warning about the more important stuff.

Additionally I turned off warnings [4244 ](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244?view=msvc-170) and [5045](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=msvc-170), 4244 because it was causing ~350 warnings alone and as such I doubt they'd all get fixed manually one by one and 5045 because it's a pretty useless warning in asar's case. 

To remove another ~90 warnings given by [9025 ](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-warning-d9025?view=msvc-170) I made use of CMake's [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/v3.24/prop_tgt/MSVC_RUNTIME_LIBRARY.html) target property. This is because if you manually specify /MT or /MD on the target_compile_options() CMake will just append them resulting in both /MT and /MD in the command line, because [from CMake's docs]

> If this property is not set then CMake uses the default value MultiThreaded$<$<CONFIG:Debug>:Debug>DLL to select a MSVC runtime library.

obviously the compiler will ignore the first and only use the second one but it will still print a warning about it. Using that specific target property avoids the duplicate command line flag and thus the warning. Note that this is why I had to add 

`cmake_policy(SET CMP0091 NEW)`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,3 +1,4 @@
+cmake_policy(SET CMP0091 NEW)
 project(Asar)
 
 cmake_minimum_required(VERSION 3.9.0)
--- a/src/asar/CMakeLists.txt
+++ b/src/asar/CMakeLists.txt
@@ -19,12 +19,19 @@ set (CMAKE_LINKER_FLAGS_DEBUG "${CMAKE_LINKER_FLAGS_DEBUG} -fsanitize=address")
 # stand-alone application and library interface
 
 macro(set_asar_shared_properties target msvc_lib_type_param enable_sanitizer)
+	# this is here because in CMake you can't modify macro parameters
+	# so we create a new variable
+	set(msvc_lib_type_var ${msvc_lib_type_param})
 	if (MSVC)
 		if (MSVC_LIB_TYPE)
-			set(msvc_lib_type_param ${MSVC_LIB_TYPE})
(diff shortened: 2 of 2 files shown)


RPGHacker:
4244 actually sounds like it could still be a useful warning in some cases. I could definitely see that preventing a few bugs that you might be searching for quite a while otherwise. Though I'm not sure how often that's actually thrown in Asar, so not sure if it's feasible to re-enable?

Atari2:
4244 is certainly useful in some cases but when it appears 
image
this many times, it's virtually impossible to distinguish between the times it's actually a concern vs the times it's harmless. Unless someone is willing to go through and check every single warning. That's why I decided to turn it off instead.

RPGHacker:
Yeah, I figured... I guess that's something for future me to clean up (aka: something that will never be cleaned up).

PR #250, opened 2022-06-23
Correct README.md

Atari2:
In #243 after I removed some changes to the PR, I didn't update the README and as such it displays wrong and/or outdated information. This fixes it and shows a correct way to include asar in other projects using CMake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,18 @@ If you'd rather not build from source, check out the [Releases](https://github.c
 Asar can also be built as a DLL. This makes it easier and faster to use in other programs (such as a sprite insertion tool). You can find documentation on the DLL API in the respective bindings (asardll.h, asar.cs, asar.py).
 
 ## Asar as a static library
-Asar can also be build as a static library. All "out-facing" functions are in interface-lib.h. This is useful for embedding Asar in other programs which don't want to use DLLs. The easiest way to add asar as a static library to your project, assuming you are using CMake (at least 3.11), is to use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) to fetch the source code, then add the following to your CMakeLists.txt:
+Asar can also be build as a static library. All "out-facing" functions are in interface-lib.h. This is useful for embedding Asar in other programs which don't want to use DLLs. The easiest way to add asar as a static library to your project is to add it as a git submodule 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,11 +14,18 @@ If you'd rather not build from source, check out the [Releases](https://github.c
 Asar can also be built as a DLL. This makes it easier and faster to use in other programs (such as a sprite insertion tool). You can find documentation on the DLL API in the respective bindings (asardll.h, asar.cs, asar.py).
 
 ## Asar as a static library
-Asar can also be build as a static library. All "out-facing" functions are in interface-lib.h. This is useful for embedding Asar in other programs which don't want to use DLLs. The easiest way to add asar as a static library to your project, assuming you are using CMake (at least 3.11), is to use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) to fetch the source code, then add the following to your CMakeLists.txt:
+Asar can also be build as a static library. All "out-facing" functions are in interface-lib.h. This is useful for embedding Asar in other programs which don't want to use DLLs. The easiest way to add asar as a static library to your project is to add it as a git submodule
```
RPGHacker, line 17: I think it should be "be built" here instead of "be build".

README.md
```diff
@@ -14,11 +14,18 @@ If you'd rather not build from source, check out the [Releases](https://github.c
...
+then add the following to your CMakeLists.txt:
 ```CMake
-target_include_directories(YourTarget PUBLIC ${asar_SOURCE_DIR}/src)
+add_subdirectory(<path-to-subdir>/src)
+get_target_property(ASAR_INCLUDE_DIR asar-static INCLUDE_DIRECTORIES)
+include_directories(${ASAR_INCLUDE_DIR})
+target_link_libraries(YourTarget PUBLIC asar-static)
 ```
-to be able to include the header files. It is also recommended to add `set(ASAR_TESTING_DISABLED TRUE)` to your CMakeLists.txt to disable building tests.
+to be able to include the header files. It is also recommended to turn off every build in target in asar except the static one using the appropriate CMake options. You will need to also make sure that your project has an Asar compatible license.
```
RPGHacker, line 28: Can we remove the "also" from "You will need to also"? The previous sentence already has an "also", and I think this sentence works fine without.

PR #254, opened 2022-07-16
Allocate thisone using new

xfix:
`endmacro` deallocates it using `delete`. Using `delete` on memory allocated with `malloc` is UB.


RPGHacker:
Were exactly was that offending delete call? I just checked, and assocar seems to use free(), so malloc should be correct here.

I did, however, find a memory leak related to thisone, which I already fixed on macro_fuckery in 62ac19989dd0ab916795bea70f39d0ea5918c01d

RPGHacker:
Okay, I found the delete. I'm not sure when it changed to a free, but yeah, free is more correct here, due to assocarr.

xfix:
https://github.com/RPGHacker/asar/blob/f2793fe0ab348dcfc07047833d51c89b79f85ff2/src/asar/macro.cpp#L90

But fair enough, it seems you have a fix for this issue that is currently not commited to `master`.

RPGHacker:
Yeah, I don't actually remember replacing that delete with a free(), but it seems I did, lol. Will have to remember cherry-picking the fix into Asar 1.9.

PR #266, opened 2022-10-07
Emacro_redefined: print where original definition is

spooonsss:


p4plus2:
Todo, fix this.  Pretty sure the build will fail, but I need a reminder for later

PR #284, opened 2023-06-21
fix BRL documented syntax

jeffythedragonslayer:


RPGHacker:
Obsoleted by https://github.com/RPGHacker/asar/commit/2b29a7d988837f819c20271322e2688ba364bb5b.

RPGHacker:
> overcome by events

What's that even supposed to mean?

RPGHacker:
But "overcome some other way" is literally one of the meanings of "obsoleted", isn't it? 🥲
At least that's how I've always used the word.

RPGHacker:
Where specifically would "where I come from" be? And how exactly is it considered rude there? It's the first time I'm hearing this, so it doesn't seem to be a universal thing, at least. I've also never heard about "OBE" before today, so it doesn't seem to be quite that widespread. I have my doubts that everyone immediately understands what it means.

Don't take this as me being insistent on sticking with my current terminology - I don't want to be rude to people if it can be avoided. But I at least need to understand where people are coming from before I change the way I speak. At least from my perspective, "obsolete" is just formal progamming talk, and if it's indeed just some regions where it's considered offensive, I don't think people in other places necessarily need to ditch the word because of that.

RPGHacker:
Alright, acknowledged. I can definitely see how "obsolete" can come of as rude in some contexts - I just don't necessarily see the word itself as rude, and in this particular context I wouldn't consider it that, since we're talking about pretty quick and trivial fix here. It'd certainly be a different story if this were a bigger pull request with a larger time investment - but I reckon this only took a few minutes at most, right?

Nevertheless, I can see where you're coming from, so I'll be more mindful.

jpotter-oms:
What exactly are you pointing out?  That link just goes to the first page of a 6502 manual.

jpotter-oms:
I'll be sure to forward it to HR ( @RPGHacker )

RPGHacker:
Acknowledged! 🙂
(Though to be fair, that book is from 1984 - when it was released, I wasn't even born yet, so it doesn't necessarily say much about how common the term is nowadays)

PR #299, opened 2024-01-29
Add support for the rest of the patchparams struct members, and add corresponding parameters to 'patch' method.

ghost:


src/asar-dll-bindings/c_sharp/asar.cs
```diff
@@ -208,6 +249,30 @@ private struct RawPatchParams
                     defines[i].contents = Marshal.StringToCoTaskMemAnsi(value);
                 }
 
+                var warningKeys = warningSettings.Keys.ToArray();
+
+                for (int i = 0; i < warningSettings.Count; i++)
+                {
+                    var warnId = warningKeys[i];
+                    var value = warningSettings[warnId];
+                    warnings[i].warnid = (byte*)Marshal.StringToCoTaskMemAnsi(warnId);
```
trillllian, line 258: don't you also need to call `FreeCoTaskMem` on these in the `finally{}` block below

ghost: Ah, good catch. Let me add that in as well for those.

PR #308, opened 2024-03-01
src/asar: fix -Wmisleading-indentation warnings

orbea:
These macros are impossible to read and cause extremely verbose `-Wmisleading-indentation` warnings.

### DIFF
--- a/src/asar/arch-spc700.cpp
+++ b/src/asar/arch-spc700.cpp
@@ -329,8 +329,15 @@ bool asblock_spc700(char** word, int numwords)
 #define vc(left1, right1, str2) if (isvc(left1, right1, str2))
 #define vv(left1, right1, left2, right2) if (isvv(left1, right1, left2, right2))
 #define w0(opcode) do { write1((unsigned int)opcode); return true; } while(0)
-#define w1(opcode, math) do { write1((unsigned int)opcode); unsigned int val=getnum_ck(math); \
-		if ((((val&0xFF00)&&(val&0x80000000)==0)||(((val&0xFF00)!=0xFF00)&&(val&0x80000000)))&&opLen!=1) asar_throw_warning(2, warning_id_spc700_assuming_8_bit); write1(val);return true; } while(0)
+#define w1(opcode, math) \
+	do { \
+		write1((unsigned int)opcode); \
+		unsigned int val=getnum_ck(math); \
+		if ((((val&0xFF00) && (val&0x80000000) == 0) || (((val&0xFF00) != 0xFF00) && (val&0x80000000))) && opLen != 1) \
+			asar_throw_warning(2, warning_id_spc700_assuming_8_bit); \
+		write1(val);\
+		return true; \
+	} while(0)
 #define w2(opcode, math) do { write1((unsigned int)opcode); write2(getnum_ck(math)); return true; } while(0)
 #define wv(opcode1, opcode2, math) do { if ((opLen == 1) || (opLen == 0 && getlen(math)==1)) { write1((unsigned int)opcode1); write1(getnum_ck(math)); } \
 																	 else { write1((unsigned int)opcode2); write2(getnum_ck(math)); } return true; } while(0)
--- a/src/asar/libstr.cpp
+++ b/src/asar/libstr.cpp
@@ -156,9 +156,50 @@ bool readfile(const char * fname, const char * basepath, char ** data, int * len
 	return true;
 }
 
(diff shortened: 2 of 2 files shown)


src/asar/libstr.cpp
```diff
@@ -156,9 +156,50 @@ bool readfile(const char * fname, const char * basepath, char ** data, int * len
...
+		while (par) { \
+			dequote(var, next, error); \
+		else { \
+			if (var=='(') par++; \
+			if (var==')') par--; \
+			if (!var) error; \
+			next; \
+		} \
+	} \
+} else if (var==')') error
```
Alcaro: I think this section should have another level of indentation. Looks like your formatter doesn't like that dequote macro.

Can't say I disagree.

src/asar/arch-spc700.cpp
```diff
@@ -329,8 +329,15 @@ bool asblock_spc700(char** word, int numwords)
 #define vc(left1, right1, str2) if (isvc(left1, right1, str2))
 #define vv(left1, right1, left2, right2) if (isvv(left1, right1, left2, right2))
 #define w0(opcode) do { write1((unsigned int)opcode); return true; } while(0)
-#define w1(opcode, math) do { write1((unsigned int)opcode); unsigned int val=getnum_ck(math); \
-		if ((((val&0xFF00)&&(val&0x80000000)==0)||(((val&0xFF00)!=0xFF00)&&(val&0x80000000)))&&opLen!=1) asar_throw_warning(2, warning_id_spc700_assuming_8_bit); write1(val);return true; } while(0)
+#define w1(opcode, math) \
+	do { \
+		write1((unsigned int)opcode); \
+		unsigned int val=getnum_ck(math); \
+		if ((((val&0xFF00)&&(val&0x80000000)==0)||(((val&0xFF00)!=0xFF00)&&(val&0x80000000)))&&opLen!=1) \
```
Alcaro: Can't say this is particularly readable even with that fix.

But it's slightly less unreadable than before, at least.

orbea: I added spaces, that makes it a little better I think?

src/asar/libstr.cpp
orbea: Moving `error` to the next line with an indent makes the compiler unhappy again... For the record I am using `gcc-13.2.1`.

src/asar/arch-spc700.cpp
Alcaro: Slightly, yes, but the main issue is still there's four or five levels of nested parentheses.

But it's an improvement, that's what counts. I vote merge this, we can worry about that specific line later. Or never.

src/asar/libstr.cpp
Alcaro: No, I don't mean just that error, I mean the entire chunk from line 195 to 202.

Alcaro (approved):

src/asar/libstr.cpp
orbea: Oh, thanks for explaining, I see what you mean now. I think I corrected it? Regardless `error` on its own line make the compiler unhappy still...

Alcaro: Yes, that's the spot. Looks good to me.

I have no strong feelings one way or the other about that specific word; if the compiler has an opinion, let's just do whatever it prefers.

It would surprise me if that's the only -Wmisleading-indentation, but maybe Asar is such a labyrinth the compiler just gives up.

src/asar/arch-spc700.cpp
orbea: Thanks for the quick review! I would prefer if any code changes besides indents and spaces were left for a followup.

p4plus2:
A lot of these won't (or already don't) apply to the 2.0, but no reason to not fix them in 1.9 in the meantime.  Thanks for the PR

If you're interested we do have a discord channel on sneslab for asar development and discussion. https://discord.com/invite/jh5Ry77N

orbea:
> A lot of these won't (or already don't) apply to the 2.0, but no reason to not fix them in 1.9 in the meantime.

Should I be making PRs against the `asar_2_beta` branch?

p4plus2:
Bug fixes and cleanup are acceptable in 1.9, but any features should belong in 2.0. But as a fair warning it is undecided whether 1.9 will see another release (though i suspect a 1.9.1 will happen)

2.0 is overall a long way from complete but initial cleanup and improvements are moving in the right direction.  I'd also estimate 2.0 is roughly 3-4x faster than 1.9 at this time if you're working on large projects.

PR #311, opened 2024-03-02
asar-tests: fix build with musl

orbea:
Trivial build fix for musl systems, this is a standard header and it should be safe to include everywhere even if the libc implicitly includes it already.
```
src/asar-tests/test.cpp: In function 'int main(int, char**)':
src/asar-tests/test.cpp:958:57: error: 'int64_t' was not declared in this scope
  958 |          test_status &= (int64_t)asar_math("1+0", &error_buffer) == 1;
      |                          ^~~~~~~
```

### DIFF
--- a/src/asar-tests/test.cpp
+++ b/src/asar-tests/test.cpp
@@ -28,7 +28,6 @@
 #	include <algorithm>
 #	include <set>
 #	include <list>
-#	include <cstdint>
 #	include <errno.h>
 
 #	if defined(_MSC_VER)
@@ -47,6 +46,8 @@
 #	include <list>
 #	include <errno.h>
 #endif
+
+#include <cstdint>
 
 #if defined(ASAR_TEST_DLL)
 #	include "asardll.h"


trillllian:
same comment as the other tests PR

ughhhhhh okay i don't know how to use github i guess. i'll cherry-pick this into 1.9 manually

orbea:
> ughhhhhh okay i don't know how to use github i guess. i'll cherry-pick this into 1.9 manually

The failure only happens with `asar_2_beta` fwiw.

trillllian:
well i suppose i better get on with deleting that code then

interestingly, we do have CI running on my raspberry pi with alpine, which uses musl, so it's weird that it didn't show up there (maybe gcc still uses glibc headers or something??)

orbea:
That is strange, I have musl `1.2.4` so maybe its a difference in the version? Regardless `int64_t` is clearly defined as requiring `stdint.h`/`cstdint`.

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdint.h.html

PR #313, opened 2024-03-11
Add missing include in asardll.c

Atari2:
This commit adds an `#include <stdlib.h>` in asardll.c This is because otherwise malloc and free functions are not defined anywhere, causing warnings (at least in msvc).

e.g.
warning C4013: 'malloc' undefined; assuming extern returning int 
warning C4013: 'free' undefined; assuming extern returning int


Alcaro (approved):
I thought you had push access to this repo

PR #323, opened 2024-07-01
build: macOS fixes

jlopez:
- **build: do not use threads on macOS builds**
- **build: fix macOS compile warnings**


trillllian:
the correct fix is to use osx specific pthread functions, instead of disabling all usage of threads. seem to be `pthread_get_stackaddr_np` and `pthread_get_stacksize_np`, though they aren't particularly well-documented. but i guess i can cherry-pick the warning fixes.

PR #350, opened 2025-02-18
Optimizations to string and assocarr

yuriks:
These give a ~17% speedup when building [sm_disassembly](https://github.com/InsaneFirebat/sm_disassembly) (lightly patched to work with 2.0.0)

```
hyperfine -w 1 --prepare "python tools/ff_file.py ../SM.sfc" --reference "C:\Users\yuriks\projects\asar\build_asar2\asar\bin\RelWithDebInfo\asar.exe --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc" "tools\asar2\asar --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc"
Benchmark 1: C:\Users\yuriks\projects\asar\build_asar2\asar\bin\RelWithDebInfo\asar.exe --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc
  Time (mean ± σ):      2.349 s ±  0.012 s    [User: 1.958 s, System: 0.385 s]
  Range (min … max):    2.327 s …  2.365 s    10 runs

Benchmark 2: tools\asar2\asar --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc
  Time (mean ± σ):      2.745 s ±  0.021 s    [User: 2.293 s, System: 0.432 s]
  Range (min … max):    2.720 s …  2.788 s    10 runs

Summary
  C:\Users\yuriks\projects\asar\build_asar2\asar\bin\RelWithDebInfo\asar.exe --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc ran
    1.17 ± 0.01 times faster than tools\asar2\asar --no-title-check --symbols=wla --symbols-path=symbols.sym src/main.asm SM.sfc
```


yuriks:
Seems like the CI failed to build. I didn't get any issues locally with VS 2022, but I'll have to look into it tomorrow.

Alcaro:
```c++
template<>
struct std::hash<string> {
	size_t operator()(const string& s) const {
		return std::hash<std::string_view>()(std::string_view(s.data(), s.length()));
	}
};
```
Since that's in std::hash, it's in namespace std, and const string& s resolves to std::string.

To fix, change to ::string.

yuriks:
Ah, good catch, that's a quick fix, so I pushed an amended commit.

yuriks:
The C# test is failing because the order in which labels are returned was different. Is the order of the labels in the API something that is expected to be kept stable, or should I just update the test? Otherwise, I had a previous implementation with `std::map` that I can use instead so they're returned in sorted order.

trillllian:
i think it would be nice to return sorted data from the api functions. but my preferred implementation would be to just call sort() on the output before returning it, it'd probably be more efficient than using a sorted map in the rest of asar.

also, it'd probably be interesting to see if using a better hashmap implementation (i think absl's one was good?) would give noticeable speedups.

trillllian:
oh and CI is failing on linux because you're missing `#include <cstring>` or something. i guess msvc is more lenient about standard includes

trillllian:
oh right, the symbol file also contains all labels, so you'd need to do the sort() [somewhere here](https://github.com/RPGHacker/asar/blob/5e0cdba7566438be20abf4591633c01d83750dd9/src/asar/main.cpp#L926) too.

i hope that that's the last one...

yuriks:
Yeah, I identified that the symbols output would change, but also didn't see there were separate tests for that. I made it sort by address rather than label name, which I think is generally what one would expect. This should be the last one, all the other uses of `.each` didn't care about ordering.

yuriks:
> also, it'd probably be interesting to see if using a better hashmap implementation (i think absl's one was good?) would give noticeable speedups.

re. this, I agree `std::unordered_map` is far from optimal, but the label assocmap is not showing up much in the profile anymore (it's mostly string parsing and copying, whereas the main cost in the previous assocmap implementation was the repeated strcmp-ing of strings), so I don't think it will make much of a difference to sub in a faster implementation.

PR #378, opened 2025-08-14
Replace outdated endif usage with endwhile in manual

yuriks:


Alcaro (approved):
we didn't do this long ago?

